PR #1, opened 2024-08-09
Add initial Spark refactor

### LINKED ISSUE #3: Investigate replacing sqoop with spark

dfsnow:
This PR represents an initial attempt to refactor the increasingly slow and unwieldy [service-sqoop-iasworld](https://github.com/ccao-data/service-sqoop-iasworld) to Spark. This refactor has numerous benefits:

- Spark is the industry standard for this type of work, while sqoop is deprecated and no longer maintained
- PySpark code is infinitely more readable and maintainable compared to the sqoop code, which is mostly bash
- PySpark extractions tend to be a full order of magnitude faster than sqoop due to the use of SQL predicates to create more discrete extract tasks
  - `ASMT_ALL WHERE taxyr IN (2023, 2024)` via sqoop = 1H30M
  - `ASMT_ALL WHERE taxyr IN (2023, 2024)` via Spark = 14M
- Job definitions are clearer and easier to maintain (via their own dedicated YAML file). They're also far more flexible e.g. only get `cur = 'Y'` values for 2023
- The output of the extract jobs is a zstd-compressed Parquet file per partition, compared to the many files currently created per partition via sqoop. This should make our Athena queries slightly faster
- Making table extracts only via indexed columns (PARID, TAXYR) may actually result in faster queries AND less load on the iasWorld database (compared to Enquire queries, which use a lot of un-indexed columns)

Note that this PR only includes the extract and partitioning parts of the refactor. I figured it would be worthwhile to break the review up into more manageable chunks. I still need to add:

- Upload to S3
- Job validation
- Logging
- Schema assurance/conformity to existing sqoop types


.env
dfsnow, line 1: This file is just a demo/template. The actual file in the repo is filled out and changes to it are ignored.

.github/workflows/build-docker.yaml
dfsnow, line 1: GitHub workflows haven't changed other than bumping the Action versions. Feel free to skip reviewing them.

README.md
dfsnow, line 1: This is just a first pass at the README. Will fill it out more as things get added. That said, it should give you a pretty good approximation of what's going on in the repo.

README.md
```diff
@@ -1 +1,95 @@
...
+Parquet files to [AWS S3](https://aws.amazon.com/s3/). The Data Department then
+queries these Parquet files using [AWS Athena](https://aws.amazon.com/athena),
+giving us a 1-1 mirror of the system-of-record for analytical queries.
+
+## Submitting jobs
+
+> [!NOTE]
+> Before attempting to submit jobs to the cluster, first make sure the Spark
+> Docker Compose stack is active by running `docker compose up -d` in the
+> repository.
```
dfsnow, line 21: This is a change compared to sqoop, which called `docker compose` via a shell script to spin up and down its cluster. I think the always-on way is preferable since then we can use the Spark web UI to view job history.

README.md
```diff
@@ -1 +1,95 @@
...
+
+`service-spark-iasworld` jobs are submitted via JSON, either as a string or
+as a file. All jobs should have the following format:
+
+```json
+{
+  "addn": {
```
dfsnow, line 28: The "job name" doesn't actually matter. It's not used in any way, just there for reference.

config/default_jobs.yaml
dfsnow, line 1: This file defines jobs via YAML, which gets converted to the JSON shown in the README. This format has a lot of benefits:

- We can define "sets" of jobs to run i.e. all weekend jobs or all jobs needed for QC reports
- Jobs only require the table name. Any other fields that aren't defined just use the defaults. This saves lots of space / reduces redundancy.
- We can use YAML templating to change lots of things at once. See `min_year`

config/default_predicates.csv
dfsnow, line 1: This file is the big brain bit that makes Spark way faster than sqoop. Here's why:

Both Spark and sqoop extract via JDBC. To do so, the divide up the data that needs to be extracted into partitions, then assign one worker to pull each partition. This works great unless your partitions are drastically different sizes.

For sqoop, we partitioned by `taxyr`, which works well for the most part, but _really_ chugs on tables where a single `taxyr` is huge (like `ASMT_ALL`).

So what can we do to divide up the range of data into smaller chunks? Partition on PIN! This file defines the partition boundaries of roughly equally sized chunks of PINs. Each one of these chunks then gets a worker, which makes extracts super fast.

> [!NOTE]
> One point of uncertainty I have here: I don't account for possibly null values of `parid`, `taxyr`, or `cur`. These aren't supposed to be nullable columns, but :shrug:.

docker-compose.yaml
```diff
@@ -0,0 +1,79 @@
...
+      - SPARK_RPC_AUTHENTICATION_ENABLED=no
+      - SPARK_RPC_ENCRYPTION_ENABLED=no
+      - SPARK_LOCAL_STORAGE_ENCRYPTION_ENABLED=no
+      - SPARK_SSL_ENABLED=no
+      - SPARK_USER=spark
+      - IPTS_HOSTNAME
+      - IPTS_PORT
+      - IPTS_SERVICE_NAME
+      - IPTS_USERNAME
+      - IPTS_TABLE
```
dfsnow, line 38: These are shared by both containers, but unfortunately I don't think there's a way to dedupe them.

run.sh
dfsnow, line 1: This is just a lil example file showing how to submit jobs from YAML. I imagine I'll expand this script so we can use it via cron jobs and the like.

src/submit_jobs.py
dfsnow, line 1: This script does all the actual work of submitting and tracking the jobs. It basically: 

1. Turns the JSON into jobs
2. Creates a single Spark session to share across all jobs
3. Runs all the extract jobs, then closes the Spark session to free memory
4. Uses PyArrow to repartition the Spark read output to a ZFS-compressed single file per partition

I tried to comment it thoroughly, so I won't use PR comments too much here. Let me know if anything is unclear.

> [!NOTE]
> P.S. @jeancochrane it would be very helpful if you could review this (and other Python modules) for idiomatic Python, since I'm usually an R guy.

src/submit_jobs.py
```diff
@@ -0,0 +1,116 @@
...
+    # chunks by PARID (where available) since it is indexed (we get fast reads)
+    predicates_csv = read_predicates(PATH_PREDICATES)
+
+    # Each session is shared across all read jobs and manages job order,
+    # credentialing, retries, etc.
+    current_datetime = datetime.now().strftime("%Y-%m-%d_%H:%M:%S")
+    session_name = f"iasworld_{current_datetime}"
+    session = SharedSparkSession(
+        app_name=session_name, password_file_path=PATH_IPTS_PASSWORD
+    )
```
dfsnow, line 64: Note that you can connect to an existing session by passing the same name, which might be helpful in a future where we want to poll the status of jobs.

src/utils/spark.py
dfsnow, line 1: This is the real bulk of the PR. Figuring out all these Spark options and the right abstractions took me awhile. I think I landed on something pretty simple and understandable, but open to suggestions!

.env
dfsnow, line 1: This file is just a demo/template. The actual file in the repo is filled out and changes to it are ignored.

.github/workflows/build-docker.yaml
dfsnow, line 1: GitHub workflows haven't changed other than bumping the Action versions. Feel free to skip reviewing them.

config/default_jobs.yaml
dfsnow, line 1: This file defines jobs via YAML, which gets converted to the JSON shown in the README. This format has a lot of benefits:

- We can define "sets" of jobs to run i.e. all weekend jobs or all jobs needed for QC reports
- Jobs only require the table name. Any other fields that aren't defined just use the defaults. This saves lots of space / reduces redundancy.
- We can use YAML templating to change lots of things at once. See `min_year`

config/default_predicates.csv
dfsnow, line 1: This file is the big brain bit that makes Spark way faster than sqoop. Here's why:

Both Spark and sqoop extract via JDBC. To do so, the divide up the data that needs to be extracted into partitions, then assign one worker to pull each partition. This works great unless your partitions are drastically different sizes.

For sqoop, we partitioned by `taxyr`, which works well for the most part, but _really_ chugs on tables where a single `taxyr` is huge (like `ASMT_ALL`).

So what can we do to divide up the range of data into smaller chunks? Partition on PIN! This file defines the partition boundaries of roughly equally sized chunks of PINs. Each one of these chunks then gets a worker, which makes extracts super fast.

> [!NOTE]
> One point of uncertainty I have here: I don't account for possibly null values of `parid`, `taxyr`, or `cur`. These aren't supposed to be nullable columns, but :shrug:.

run.sh
dfsnow, line 1: This is just a lil example file showing how to submit jobs from YAML. I imagine I'll expand this script so we can use it via cron jobs and the like.

src/utils/spark.py
dfsnow, line 1: This is the real bulk of the PR. Figuring out all these Spark options and the right abstractions took me awhile. I think I landed on something pretty simple and understandable, but open to suggestions!

I tried to document it thoroughly via docstrings and comments, rather than PR comments, since I want this to be easy to grok next time we come back and work on it.

src/utils/spark.py
```diff
@@ -0,0 +1,223 @@
...
+            df = df.filter(filter)
+
+        (
+            df.write.mode("overwrite")
+            .option("compression", self.session.initial_compression)
+            .partitionBy(partitions)
+            .parquet(self.initial_dir)
+        )
+
+    def repartition(self) -> None:
```
dfsnow, line 188: This method is called after _all_ the Spark extract jobs have run. If you try to call it after each job it will crash the server. Spark doesn't like to free memory, neither does PyArrow. They fight to the death and chungus loses.

Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM bitnami/spark:3.5.1
+SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
+
+# Switch to root user so we can install jq and yq for parsing JSON/YAML
+USER root
+RUN mkdir -p /tmp/python && \
+    mkdir -p /var/lib/apt/lists/partial && \
+    apt-get update && \
+    apt-get install --no-install-recommends -y wget=1.21.3-1+b2 jq=1.6-2.1 && \
+    wget https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64 -q -O /usr/bin/yq && \
```
jeancochrane: **[Question, non-blocking]** What's your reasoning for pinning `jq` and `wget` to specific versions but not pinning `yq`?

README.md
```diff
@@ -1 +1,95 @@
...
+
+### Field definitions
+
+- `table_name (required)` - Name of the iasWorld table to extract, must be
+  prefixed with `iasworld.`.
+- `min_year (optional)` - Minimum tax year to extract from the table.
+  Default is `1999`.
+- `max_year (optional)` - Maximum tax year (inclusive) to extract from the
+  table. To extract a single year, set `min_year` and `max_year` to the same
+  value. Default is the current year.
```
jeancochrane: **[Question, non-blocking]** Based on the single-year instructions I take it the `min_year` parameter is also inclusive, yeah?

README.md
jeancochrane: **[Praise]** Excellent docs, thank you!

run.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+# Example command to execute test jobs in the Spark cluster
+# First run `docker compose up` in the repo root
```
jeancochrane: **[Nitpick, non-blocking]** Hopefully readers will understand what you're getting at, but in case we want to be more explicit:

```suggestion
# First run `docker compose up -d` in the repo root
```

src/utils/helpers.py
```diff
@@ -0,0 +1,95 @@
+import csv
+import itertools
+import json
+from argparse import Namespace
+from pathlib import Path
+
+
+def construct_predicates(
+    csv_lines: list[list[str]], taxyr: int | list[int] | None = None
```
jeancochrane: **[Question, non-blocking]** As `src.submit_jobs.main` is currently written, `taxyr` will always be either `NoneType` or `list[int]`, right? Are we preserving the option of `int` in anticipation of a future where that might be more convenient than a one-element list?

src/utils/helpers.py
```diff
@@ -0,0 +1,95 @@
...
+        taxyr: A single tax year, a list of tax years, or None. If provided,
+            will be appended to the PARID predicates. Defaults to None.
+
+    Returns:
+        A list of SQL predicate strings used to divide a table into chunks.
+    """
+    taxyr_list = [taxyr] if isinstance(taxyr, int) else taxyr
+    if taxyr_list:
+        predicates = [
+            f"parid >= '{start}' AND parid <= '{end}' AND taxyr = {year}"
```
jeancochrane: **[Question, non-blocking]** Is `taxyr` stored as an int in the database that this predicate is referencing, or are single quotes just optional in this case?

src/utils/helpers.py
```diff
@@ -0,0 +1,95 @@
...
+        A list of start and end PARIDs used to divide a table. The third value
+        is the rough number of PARIDs in the batch.
+    """
+    pred_path = Path(path)
+    with open(pred_path.resolve().as_posix(), mode="r") as file:
+        csv_reader = csv.reader(file)
+        next(csv_reader)
+        csv_lines = [row for row in csv_reader]
+
+    return csv_lines
```
jeancochrane: **[Thought, non-blocking]** Since the predicate reading logic in this function seems to be a pretty simple operation, and its output only seems to be used by `construct_predicates`, I wonder if we could simplify things by just having `construct_predicates` accept a path to a predicates file and take responsibility for reading the predicates itself. I don't have strong feelings about it, just keeping an eye out for ways to keep the logic as simple as we can.

src/utils/spark.py
```diff
@@ -0,0 +1,223 @@
...
+    """
+
+    def __init__(
+        self,
+        session: SharedSparkSession,
+        table_name: str,
+        taxyr: int | list[int] | None,
+        cur: str | list[str] | None,
+        predicates: list[str] | None,
+        use_partitions: bool,
```
jeancochrane: **[Thought, non-blocking]** Not a big deal since it's one extra param that directly maps onto a job config param, but it seems like the value of `use_partitions` could be derivable from the values of `cur` and `taxyr` in case we want to simplify things by getting rid of a parameter.

src/utils/spark.py
```diff
@@ -0,0 +1,223 @@
...
+        # partitions by taxyr but not cur, and visa-versa
+        if self.use_partitions and (
+            (self.taxyr is None) != (self.cur is None)
+        ):
+            raise ValueError(
+                (
+                    f"Error for table {self.table_name}: "
+                    "both 'taxyr' and 'cur' must be set if one is set."
+                )
+            )
```
jeancochrane: **[Question, non-blocking]** Given that both `cur` and `taxyr` have default values that are set in `main`, is it possible to configure job parameters that accidentally set one of `taxyr` or `cur` but not the other? Or is this just defensive programming for a future where we might be instantiating `SparkJob` differently (e.g. via an API)? If it's the latter, is there a good reason we don't want to allow a caller to set one or the other of these params but not both?

src/utils/helpers.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def strip_table_prefix(table_name: str) -> str:
+    """Removes the 'iasworld.' prefix from a table name."""
+    if table_name.startswith("iasworld."):
+        return table_name.split(".", 1)[1]
+    return table_name
+
+
+def load_job_config(args: Namespace) -> dict:
```
jeancochrane: **[Suggestion, non-blocking]** I have a preference for functions that explicitly name (and type hint) their arguments rather than accepting one object with an arbitrary number of attributes, since the latter design forces readers to read the implementation of the function in order to know which attributes the function ends up using. It's not a big deal though, and I think people disagree on this!

src/submit_jobs.py
```diff
@@ -0,0 +1,116 @@
...
+        jobs.append(spark_job)
+
+    # Stop the Spark session once all JDBC reads are complete. This is CRITICAL
+    # as it frees all memory from the cluster for use in the repartition step
+    session.spark.stop()
+
+    # Use pyarrow to condense the many small Parquet files created by the reads
+    # into single files per Hive partition
+    for config in jobs:
+        config.repartition()
```
jeancochrane: **[Nitpick, non-blocking]** I find `config` to be a bit of a confusing name in this case since it was being used earlier in the function to describe a different object. Is there a reason not to just use `spark_job` again as we do on line `100`?

```suggestion
    for spark_job in jobs:
        spark_job.repartition()
```

src/submit_jobs.py
```diff
@@ -0,0 +1,116 @@
...
+            cur=cur,
+            predicates=predicates,
+            use_partitions=use_partitions,
+            initial_dir=PATH_INITIAL_DIR,
+            final_dir=PATH_FINAL_DIR,
+        )
+
+        # Run the Spark read job. Each job will create n_predicates X n_years
+        # files in the target/initial/ dir, assuming predicates are enabled
+        spark_job.read()
```
jeancochrane: **[Question, non-blocking]** If I'm understanding correctly, this blocks until the full matrix of predicate reads is complete, is that right? If so, it's making me wonder if there's possibly even more performance that we could eke out by parallelizing the jobs themselves too, such that workers within a job could move onto the next job if all predicate reads have been picked up but not yet completed. Highly speculative, so not something I expect us to investigate now!

src/utils/spark.py
```diff
@@ -0,0 +1,223 @@
...
+        Translates the taxyr and cur values into SQL used to filter/limit the
+        values read from the table.
+        """
+        if self.taxyr is None:
+            return None
+
+        if isinstance(self.taxyr, list):
+            filter = f"taxyr IN ({', '.join(map(str, self.taxyr))})"
+        else:
+            filter = f"taxyr = {self.taxyr}"
```
jeancochrane: **[Question, non-blocking]** Maybe i'm missing something, but aren't we already filtering by `taxyr` via the SQL expressions that comprise the predicates matrix?

jeancochrane (approved):
Really awesome work here! I have no major complaints, just a few questions and small suggestions below.

From a big-picture design perspective, one thing I notice is that the `SparkJob` class is currently designed around the assumption that `cur`, `parid`, and `taxyr` are the only options for our filters/predicates/partitions. If we ever need to change that assumption, I expect we'll need to rewrite most of the `SparkJob` methods to accommodate new or changed options. I think this is probably fine, since we can always refactor the class to make the filter/predicate/partition options more general, but it's one way in which the abstraction as it's currently designed isn't maximally flexible.

I also found the distinction between filters, predicates, and partitions to be a bit confusing, since they all seem to be different kinds of logic conditions, but I think I finally grok it after reading the code. Let me know if this matches your understanding:

* **Predicates** are logic conditions that are applied to read queries _during_ query execution against the iasWorld database, and are used to construct a matrix of mutually exclusive queries that can run in parallel. 
    * Predicates are only enabled if `use_predicates` is set, in which case they are always `parid` and `taxyr`. Users can override the `parid` predicates by editing the predicates CSV file, and can override the `taxyr` predicates by setting `min_year` and/or `max_year` in the job config.
* **Filters** are logic conditions that are applied to the result set _after_ query execution against the iasWorld database.
    * Filters are effectively only enabled if `use_partitions` is set, since otherwise the `main` function will set the `years` and `cur` variables to `None` rather than reading them from the job config. When enabled, filters are always `cur` and `taxyr`. Users can override the values for these filters by setting `min_year`, `max_year`, and/or `cur` in the job config.
* **Partitions** are logic conditions that are applied to the result set after filtering in order to split up the output files and index the data.
    * Partitions are only enabled if `use_partitions` is set, in which case they are always `taxyr` and `cur`. Users can override the values for these filters by setting `min_year`, `max_year`, and/or `cur` in the job config.

There are probably some opportunities for simplification here, even outside of an effort to make these conditions more flexible -- for example, the fact that the filters are enabled by the `use_predicates` job config option is a bit confusing, as is the way that  the script uses `taxyr` as both a predicate and a filter. Still, I don't think any of this is super important to worry about right now!

Dockerfile
dfsnow: Being dumb and forgetful mostly. Fixed in ca7e883!

README.md
dfsnow: Indeed! Clarified in b249e95.

src/utils/helpers.py
dfsnow: Just leftovers from when I initially passed a `taxyr` value directly from the job definition. I think it's worth leaving in just to be defensive.

src/utils/helpers.py
dfsnow: `taxyr` is indeed always an int in all iasWorld tables that I know of.

src/utils/helpers.py
dfsnow: I originally did this to avoid reading the CSV on every loop, but I reworked predicates entirely in fd7ffdc. Now this function expects a CSV path that can be specified per job which contains a column, start, and end value. These values are used to create arbitrary SQL BETWEEN expressions. I also removed the interaction with taxyr for simplicity.

src/utils/helpers.py
dfsnow: Changed! See https://github.com/ccao-data/service-spark-iasworld/pull/1#discussion_r1714303698.

src/utils/spark.py
dfsnow: This was just defensiveness on my part, but I redid partitioning in 3f3c5fb such that partitions are just automatically created if `taxyr` or `cur` are set, since there _are_ tables where there's a `cur` column but no `taxyr` (SALES).

src/utils/spark.py
dfsnow: Good call, fixed in 3f3c5fb! Partitions are now just automatically created if either `cur` or `taxyr` is set.

src/submit_jobs.py
dfsnow: Fixed in 79cfdd5! I also renamed some other stuff for clarity.

src/utils/helpers.py
dfsnow: Good call! Fixed in dcce2ac.

src/utils/spark.py
dfsnow: I simplified the filter setup a bit in 4090211. Like partitions, the filter is now just constructed from any passed `taxyr` and `cur` values. I also removed the interaction of predicates and `taxyr`.

dfsnow:
Thanks for the great feedback @jeancochrane! I managed to simplify things quite a bit I think. Ready for one more round of review.

> From a big-picture design perspective, one thing I notice is that the `SparkJob` class is currently designed around the assumption that `cur`, `parid`, and `taxyr` are the only options for our filters/predicates/partitions. If we ever need to change that assumption, I expect we'll need to rewrite most of the `SparkJob` methods to accommodate new or changed options. I think this is probably fine, since we can always refactor the class to make the filter/predicate/partition options more general, but it's one way in which the abstraction as it's currently designed isn't maximally flexible.

I'm okay with this. I think it's really unlikely that we'll need to change it, and if we do then it shouldn't be too hard to rewrite. It's really simple (now) and we should keep it that way. That said, I rewrote the predicates part to be column agnostic, see below.

> I also found the distinction between filters, predicates, and partitions to be a bit confusing, since they all seem to be different kinds of logic conditions, but I think I finally grok it after reading the code.

Your understanding of these items is mostly correct, but I agree that they were really confusing. So, I scrapped the existing logic for something much simpler:

- **Predicates** are now entirely separate from `parid` and `taxyr`. They're just arbitrary SQL BETWEEN expressions loaded from a file for each job. This means we could load an arbitrary set of predicates per table. For example `SALES` could be divided up by `saledt` instead of `parid`.
- **Filters** and **partitions** are just automatically enabled as long as either `taxyr` or `cur` is not null. For example, `min_year=2021`, `max_year=2023` and `cur='Y'` will yield all data between 2021 and 2023 (inclusive), where `cur='Y'`, output to files partitioned by both `taxyr` and `cur`. If both the year values and `cur` are falsey, no filtering or partitioning is performed (we just grab the whole table).
 
> There are probably some opportunities for simplification here, even outside of an effort to make these conditions more flexible -- for example, the fact that the filters are enabled by the `use_predicates` job config option is a bit confusing, as is the way that the script uses `taxyr` as both a predicate and a filter. Still, I don't think any of this is super important to worry about right now!

Done! I think I've simplified quite a bit. Let me know what you think!

src/submit_jobs.py
dfsnow: Yep that's right. That's a good idea! Added in fa34a59. It doesn't seem to have much impact on performance, but it also doesn't seem to hurt so 🤷 .

README.md
```diff
@@ -1 +1,115 @@
...
+
+### Field definitions
+
+- `table_name (required)` - Name of the iasWorld table to extract, must be
+  prefixed with `iasworld.`.
+- `min_year (optional)` - Minimum tax year (inclusive) to extract from the
+  table. Defaults to `1999`.
+- `max_year (optional)` - Maximum tax year (inclusive) to extract from the
+  table. To extract a single year, set `min_year` and `max_year` to the same
+  value. Defaults to the current year.
```
jeancochrane: **[Suggestion, non-blocking]** On first read, I wasn't clear whether `null` would affect these two columns in the same way it affects `cur` and `predicates_path`:

```suggestion
- `min_year (optional)` - Minimum tax year (inclusive) to extract from the
  table. Set to `null` in a job definition to ignore this column when filtering
  and partitioning. Defaults to `1999`.
- `max_year (optional)` - Maximum tax year (inclusive) to extract from the
  table. To extract a single year, set `min_year` and `max_year` to the same
  value. Set to `null` in a job definition to ignore this column when filtering
  and partitioning. Defaults to the current year.
```

src/submit_jobs.py
```diff
@@ -0,0 +1,126 @@
...
+            taxyr=years,
+            cur=cur,
+            predicates=predicates,
+            initial_dir=PATH_INITIAL_DIR,
+            final_dir=PATH_FINAL_DIR,
+        )
+
+        # Run the Spark read job. Each job will create N predicates
+        # files in the target/initial/ dir, assuming predicates are enabled
+        jobs.append(spark_job)
```
jeancochrane: **[Nitpick, non-blocking]** I think this comment is out of date, right?

```suggestion
        # Register the job so that we can run them all in parallel once
        # they have all been registered. Each job will create N predicates
        # files in the target/initial/ dir, assuming predicates are enabled
        jobs.append(spark_job)
```

src/utils/helpers.py
```diff
@@ -0,0 +1,62 @@
...
+    return table_name
+
+
+def load_job_definitions(json_file: str, json_string: str) -> dict:
+    """
+    Loads job definition(s) from a JSON file or a JSON string.
+
+    Args:
+        json_file: String path to
+            either `json_file` or `json_string` value, but not both.
```
jeancochrane: **[Question, non-blocking]** Are we missing a couple of lines here? I can't quite scan what this sentence is trying to say, and the `Args` section seems to be missing an entry for `json_string`.

src/utils/helpers.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def strip_table_prefix(table_name: str) -> str:
+    """Removes the 'iasworld.' prefix from a table name."""
+    if table_name.startswith("iasworld."):
+        return table_name.split(".", 1)[1]
+    return table_name
+
+
+def load_job_definitions(json_file: str, json_string: str) -> dict:
```
jeancochrane: **[Nitpick, non-blocking]** Since these parameters are loaded from `argparse` arguments with `nargs` set to `?`, I think they should fall back to a [default](https://docs.python.org/3/library/argparse.html#default) of `NoneType`, right?

```suggestion
def load_job_definitions(json_file: str | None, json_string: str | None) -> dict:
```

README.md
```diff
@@ -1 +1,115 @@
...
+### Predicates, filtering, and partitioning
+
+- **Predicates** are SQL statements used to chunk a table during reads
+  against the iasWorld database. The statements define mutually exclusive
+  queries that run in parallel (in order to speed up query execution).
+
+  Predicates are defined via a file of SQL statements in the `config/`
+  directory, then passed to each table job via a file path.
+- **Filters** are logic conditions applied to a query as part of the initial
+  read. They're effectively a SQL WHERE clause, and apply before each read.
```
jeancochrane: **[Question, non-blocking]** I originally thought "as part of the initial read" meant filters would be applied _during_ the read, but I think you mean they're applied _after_ the read, is that right? That is, filters are applied to the dataframe that results from the Spark read, whereas predicates are used to chunk the read itself?

config/default_predicates.sql
jeancochrane, line 1: **[Question, non-blocking]** Nice refactor of this file! So did you determine that the additional `taxyr` predicates weren't helpful for speeding things up and reducing database load? Given that `taxyr` is a primary key in most tables I assumed it would be important for improving query performance when we only want a subset of years, but if it doesn't make a difference then I agree that this new format makes things much simpler.

jeancochrane (approved):
Awesome refactor! I agree that the new design is much simpler. A couple questions below, but nothing that I consider blocking. Feel free to re-request my review if you end up changing anything major, and otherwise feel free to merge once you've resolved our conversations to your liking.

README.md
dfsnow: Filters are actually composed into each SQL query made to the DB using "predicate pushdown." That means they're executed as a part of each read, rather than after. I clarified this a bit in e41f8e608989a1ad373069bbfa2464a30dee0bd0.

src/utils/helpers.py
dfsnow: Woops! Good catch. Fixed in 015737f620165f8d49a053e438ccb183790d7e4f.

config/default_predicates.sql
dfsnow: I did some testing and it seems like the addition of `taxyr` in the predicates has basically no effect outside of spawning more job tasks. The only exception is the super huge tables like `ASMT_ALL`, which each group of PINs in a BETWEEN predicate could be hundreds of thousands of rows. I'm fine with the performance <> simplicity tradeoff here though. If it becomes an issue we can make a dedicated predicates file/handler for certain tables.

PR #4, opened 2024-08-13
Add Spark job S3 upload method

dfsnow:
This PR adds an `upload()` method to `SparkJob` that takes Hive-partitioned table extraction results and uploads them to S3.


src/utils/spark.py
```diff
@@ -107,6 +116,19 @@ def __init__(
...
+        """
+        desc = [f"{self.table_name}"]
+        if self.taxyr:
+            min_year, max_year = self.taxyr[0], self.taxyr[-1]
+            desc.append(f"taxyr=[{min_year}, {max_year}]")
+        if self.cur:
+            desc.append(f"cur=[{', '.join(self.cur)}]")
+
+        return ", ".join(desc)
+
```
dfsnow, line 132: Just reordered this to be in alphabetical order with the other methods. No actual changes.

secrets/AWS_CREDENTIALS_FILE
dfsnow, line 1: This is just a boilerplate long-lived AWS credential file (though actually filled out in the repo). The benefit of passing it this way vs just mounting the `~/.aws/credentials` dir is that this container can use a set of credentials specific to its function.

That said, this feels pretty jank overall. I think we need to move toward a dedicated secret store by the end of the year. Open to any suggestions though.

dfsnow, line 1: This is just a boilerplate long-lived AWS credential file (though actually filled out in the repo). The benefit of passing it this way vs just mounting the ~/.aws/credentials dir is that this container can use a set of credentials specific to its function.

That said, this feels pretty jank overall. I think we need to move toward a dedicated secret store by the end of the year. Open to any suggestions though.

jeancochrane: **[Question, non-blocking]** Agreed that this is the right call for now, but that we should schedule some time to overhaul our credential management next quarter. Are we planning to reuse the credentials for the exisitng sqoop process, or does this Spark setup require a modified set of permissions?

src/utils/helpers.py
```diff
@@ -2,6 +2,26 @@
...
+            relative to the Docker container.
+    """
+    dir_path = Path(dir)
+    if not dir_path.is_dir():
+        raise ValueError(f"The path {dir} is not a valid directory.")
+
+    for item in dir_path.iterdir():
+        if item.is_dir():
+            clear_directory(item)  # Recursively clear
+            item.rmdir()
```
jeancochrane: **[Question, non-blocking]** Is there a reason to recursively clear subdirectories rather than use [`shutil.rmtree`](https://docs.python.org/3/library/shutil.html#shutil.rmtree)?

src/utils/spark.py
```diff
@@ -220,3 +229,59 @@ def repartition(self) -> None:
...
+            ):
+                for obj in page.get("Contents", []):
+                    s3_files_to_delete.add(obj["Key"])
+
+        if s3_files_to_delete:
+            delete_objects = [{"Key": f"{key}"} for key in s3_files_to_delete]
+            self.session.s3_client.delete_objects(
+                Bucket=self.session.s3_bucket,
+                Delete={"Objects": delete_objects},
+            )
```
jeancochrane: **[Thought, non-blocking]** It strikes me that if we delete the existing data and then run into a problem during the upload to replace it, we might end up in a situation where we wake up in the morning and find there's no data in the bucket. Is this a problem? We could be more defensive by uploading to a different prefix, clearing the old prefix, and then moving the fresh upload to the old prefix, but I'm not sure yet if the added complexity is worth it -- I'll leave it up to you to decide.

jeancochrane (approved):
Very nice work here, I like the simplicity of the `upload()` method!

secrets/AWS_CREDENTIALS_FILE
dfsnow: We can just re-use the existing credentials since they're already permissioned for exactly (and only) what this job needs to do.

src/utils/helpers.py
dfsnow: Yeah, me being dumb again. Fixed in 3e82cca.

src/utils/spark.py
dfsnow: Good thinking! In bb103cc I used a set diff such that only files not in the local dir are deleted from the remote. That way we delete orphan partitions/files, then overwrite the rest of the files via upload.

src/utils/helpers.py
jeancochrane: Nice! This may be intentional, but heads up that (by my reading) the old version of `clear_directory` would preserve the existence of the empty top-level directory itself, while `shutil.rmtree` will delete everything, including the empty directory. I'm not sure if that matters, so disregard if not!

dfsnow: Tis fine, as the top level dir is re-created on every Spark read.

PR #5, opened 2024-08-14
Add GitHub workflow dispatch method

dfsnow:
This PR ports and translates the GitHub workflow dispatch shell script from sqoop into a simple Python method for Spark. The workflow dispatch is called at the end of a session once all jobs and uploads are complete.

[Here's](https://github.com/ccao-data/data-architecture/actions/runs/10394017482) a successfully triggered workflow.


.env
```diff
@@ -2,7 +2,8 @@ AWS_S3_BUCKET=
 AWS_S3_PREFIX=
 CCAO_REGISTRY_URL=ghcr.io/ccao-data
 GH_APP_ID=
-GH_PEM_PATH=
```
dfsnow, line 5: Moved this to within Python since it never actually changes in the container.

src/utils/helpers.py
```diff
@@ -65,3 +105,31 @@ def load_job_definitions(
...
+        predicates = [line.strip() for line in file.readlines()]
+
+    return predicates
+
+
+def strip_table_prefix(table_name: str) -> str:
+    """Removes the 'iasworld.' prefix from a table name."""
+    if table_name.startswith("iasworld."):
+        return table_name.split(".", 1)[1]
+    return table_name
```
dfsnow, line 67: Just putting these in alphabetical order, no actual changes to content.

src/utils/spark.py
```diff
@@ -71,6 +87,21 @@ def __init__(self, app_name: str, password_file_path: str) -> None:
 
         self.spark = SparkSession.builder.appName(self.app_name).getOrCreate()
 
+    def run_gh_workflow(self):
```
jeancochrane: **[Thought, non-blocking]** Two thoughts on the design of this method, neither of which are particularly urgent:

1. Does it actually make sense for the `SharedSparkSession` object to own this logic? Up until now, `SharedSparkSession` has owned responsibility managing our connection to the Spark cluster; this change also gives it responsibility for managing our connection to GitHub, which feels to me like a distinct concern. Two ways we could think about separating these concerns:  
    1. Make this a bare, non-class function that takes an app ID, PEM path, repo, and workflow as arguments; keep things simple and function-oriented unless/until we decide we need to do more things with a GitHub connection such that it makes sense to factor out the connection management
    2. Match the design of `SharedSparkSession` and `SparkJob` by designing `SharedGitHubSession` and `GitHubJob` classes to split responsibility for connection/job management
2. The name `run_gh_workflow` doesn't actually make it clear _which_ workflow will run, and I could imagine us wanting to trigger multiple workflows in the future. Two options for clarity and future-proofing:
    1. Call the method something more specific like `run_gh_workflow_tests`
    2. Factor out the repo and workflow names into arguments, such that it can be called like `session.run_gh_workflow("data-architecture", "test-dbt-models")`

I don't feel super strongly about any of this, and ultimately I would be fine with us going with the current design and keeping an eye out for smells that we need more general abstractions. Up to you!

src/utils/spark.py
```diff
@@ -55,6 +65,12 @@ def __init__(self, app_name: str, password_file_path: str) -> None:
             f"{self.ipts_service_name}"
         )
 
+        # GitHub credentials for launching a target workflow once all jobs
+        # are finished running
+        self.gh_app_id = os.getenv("GH_APP_ID")
+        self.gh_repo = os.getenv("GH_REPO")
+        self.gh_workflow = os.getenv("GH_WORKFLOW")
```
jeancochrane: **[Question, non-blocking]** What's the intuition for the distinction between variables that are loaded from the environment vs. from the arguments passed to the constructor method? I find it a little bit confusing that we read class attributes from both places, but I could be convinced that it's fine if there's a good reason to do so. Otherwise, my preference would be for the classes to accept all of their attributes via constructor params, such that the responsibility falls on the calling code in `main()` to determine where to load these values.

src/utils/helpers.py
```diff
@@ -1,33 +1,73 @@
...
+
+def dispatch_gh_worfklow(gh_jwt: str, gh_repo: str, gh_workflow: str) -> None:
+    """
+    Dispatch a GitHub Actions workflow using the GitHub API.
+
+    Args:
+        gh_jwt: The JWT token for authentication with the GitHub API. Returned
+             by the `create_jwt_token` function.
+        gh_repo: API URL for the target repository containing a workflow.
+        gh_workflow: Workflow YAML file, relative to `.github/workflows`.
```
jeancochrane: **[Question, non-blocking]** Does this param require a file extension (`.yaml`)?

src/utils/helpers.py
```diff
@@ -1,33 +1,73 @@
...
+
+    response = requests.post(gh_tokens_url, headers=create_headers(gh_jwt))
+    gh_token = response.json()["token"]
+
+    data = {"ref": "master"}
+    response = requests.post(
+        f"{gh_repo}/actions/workflows/{gh_workflow}/dispatches",
+        headers=create_headers(gh_token),
+        json=data,
+    )
```
jeancochrane: **[Suggestion, non-blocking]** It might behoove us to throw some [`raise_for_status()`](https://3.python-requests.org/user/quickstart/#response-status-codes) calls in here so that we can get ahead of any HTTP errors. I don't think we need to go too deep on logging just yet, but promptly raising HTTP errors before we run into KeyErrors or JSON formatting errors seems like low-hanging fruit.

jeancochrane (approved):
Nice work!

src/utils/spark.py
dfsnow: I am dumb and should've just gone with option 2. Reworked in 9662859!

src/utils/spark.py
dfsnow: The distinction I've made so far is that anything that loads from a file is loaded via argument, while anything that loads from the environment is read on init. I mainly didn't want classes with a billion constructor params, but happy to refactor if you feel strongly!

src/utils/helpers.py
dfsnow: It does! But this is obviated by the class-based refactor I just did.

src/utils/helpers.py
dfsnow: I'm about to handle logging and raising errors in the next PR, so going to put it off until then!

src/utils/spark.py
jeancochrane: Eh, let's leave it for now, and we can easily refactor later if it becomes a pain!

dfsnow:
@jeancochrane This is ready for one more look. You were right that a dedicated class is much simpler!

.github/workflows/build-docker.yaml
```diff
@@ -18,7 +18,7 @@ jobs:
       - name: Setup variables
         run: |
           # shellcheck disable=SC2129
-          echo "IMAGE_TAG=$(if [ ${{ github.ref_name }} == master ]; then echo latest; elif [ ${{ github.event_name }} == release ]; then echo ${{ github.ref_name }}; else echo staging; fi)" >> "$GITHUB_ENV"
```
dfsnow, line 21: Unrelated fix I'm sneaking into this PR. We had the wrong default branch name and so weren't creating any `latest` tags.

jeancochrane (approved):
Good to go!

PR #6, opened 2024-08-16
Add logging, exception handling, and AWS client class

dfsnow:
This PR adds logging and error handling using the log4j driver pulled from the Spark session context. I chose to use this logger rather than standard Python logging because I want to capture the Spark output and intersperse it with Python logging.

This PR also adds a new `AWSClient` class that can trigger Glue jobs and upload finished log files to CloudWatch. I refactored the GitHub session class to more closely match the AWS one.

Here's an example [log output in CloudWatch](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/$252Fccao$252Fjobs$252Fspark/log-events/iasworld_2024-08-16_06-56-25).

### DIFF
--- a/src/submit_jobs.py
+++ b/src/submit_jobs.py
@@ -4,7 +4,11 @@ from datetime import datetime, timedelta
 
 from joblib import Parallel, delayed
 from utils.aws import AWSClient
-from utils.helpers import load_job_definitions, load_predicates
+from utils.helpers import (
+    create_fallback_logger,
+    load_job_definitions,
+    load_predicates,
+)
 from utils.github import GitHubClient
 from utils.spark import SharedSparkSession, SparkJob
 
@@ -51,26 +55,37 @@ def parse_args():
     return parser.parse_args()
 
 
-def main():
+def submit_jobs(
+    app_name: str,
+    log_file_path: str,
+    json_file: str | None = None,
+    json_string: str | None = None,
+) -> None:
+    """
+    Submit Spark jobs for iasWorld data extraction and upload.
+
+    Args:
+        app_name: Name of the job batch. Shown in the Spark UI and used as
+            the logstream name in CloudWatch.
+        log_file_path: String path to the log file for this job.
+        json_file: Path to a JSON file containing job configuration(s).
+        json_string: JSON string containing job configuration(s).
+    """
     # Get the job definition(s) from the argument JSON
     time_start = time.time()
-    args = parse_args()
-    job_definitions = load_job_definitions(
-        json_file=args.json_file, json_string=args.json_string
-    )
+    job_definitions = load_job_definitions(json_file, json_string)
 
     # Each session is shared across all read jobs and manages job order,
     # credentialing, retries, etc.
-    current_datetime = datetime.now().strftime("%Y-%m-%d_%H:%M:%S")
-    session_name = f"iasworld_{current_datetime}"
     session = SharedSparkSession(
-        app_name=session_name,
+        app_name=app_name,
+        log_file_path=log_file_path,
         password_file_path=PATH_IPTS_PASSWORD,
     )
 
     # Perform some startup logging before entering the main job loop
     table_names = [i.get("table_name") for i in job_definitions.values()]
-    session.logger.info(f"Starting Spark session {session_name}")
+    session.logger.info(f"Starting Spark session {app_name}")
     session.logger.info(f"Extracting tables: {', '.join(table_names)}")
 
     # For each Spark job, get the table structure based on the job definition
@@ -166,13 +181,21 @@ def main():
     time_duration = str(timedelta(seconds=(time_end - time_start)))
     session.logger.info(f"Total extraction duration was {time_duration}")
 
-    # Upload final logs to CloudWatch
-    aws.upload_logs_to_cloudwatch(
-        log_group_name="/ccao/jobs/spark",
-        log_stream_name=session.app_name,
-        log_file_path=session.log_file_path,
-    )
-
 
 if __name__ == "__main__":
-    main()
+    current_datetime = datetime.now().strftime("%Y-%m-%d_%H:%M:%S")
+    app_name = f"iasworld_{current_datetime}"
+    log_file_path = f"/tmp/logs/{app_name}.log"
+    fallback_logger = create_fallback_logger(log_file_path)
+    args = parse_args()
+    try:
+        submit_jobs(app_name, log_file_path, args.json_file, args.json_string)
+    except Exception as e:
+        fallback_logger.error(e)
+
+    aws = AWSClient(logger=fallback_logger)
+    aws.upload_logs_to_cloudwatch(
+        log_group_name="/ccao/jobs/spark",
+        log_stream_name=app_name,
+        log_file_path=log_file_path,
+    )
--- a/src/submit_jobs.py
+++ b/src/submit_jobs.py
@@ -136,11 +136,16 @@ def main():
     new_local_files = []
     for job in jobs:
         job_upload_results = job.upload(aws)
-        new_local_files.append(job_upload_results)
+        new_local_files.extend(job_upload_results)
 
     # If any jobs uploaded never-seen-before files, trigger a Glue crawler
-    if any(new_local_files):
-        session.logger.info("New partitions uploaded, triggering Glue crawler")
+    if new_local_files:
+        session.logger.info(
+            (
+                f"{len(new_local_files)} previously unseen files uploaded to S3, "
+                "triggering Glue crawler"
+            )
+        )
         aws.run_and_wait_for_crawler("ccao-data-warehouse-iasworld-crawler")
 
     # Trigger a GitHub workflow to run dbt tests once all jobs are complete
--- a/src/submit_jobs.py
+++ b/src/submit_jobs.py
@@ -151,9 +151,12 @@ def main():
         workflow="test_dbt_models.yaml",
     )
 
-    # Print job descriptions for extracted tables and total elapsed time
-    job_descriptions = list(map(lambda j: j.get_description(), jobs))
-    session.logger.info(f"Extracted tables: {' | '.join(job_descriptions)}")
+    # Print table names for extracted tables and total elapsed time. The
+    # get_description() call is to print the description of each table to logs
+    session.logger.info(f"Extracted tables: {', '.join(table_names)}")
+    for job in jobs:
+        job.get_description()
+
     time_end = time.time()
     time_duration = str(timedelta(seconds=(time_end - time_start)))
     session.logger.info(f"Total extraction duration was {time_duration}")
--- a/src/submit_jobs.py
+++ b/src/submit_jobs.py
@@ -3,9 +3,9 @@ import time
 from datetime import datetime, timedelta
 
 from joblib import Parallel, delayed
-from utils.aws import SharedAWSClient
+from utils.aws import AWSClient
 from utils.helpers import load_job_definitions, load_predicates
-from utils.github import SharedGitHubSession
+from utils.github import GitHubClient
 from utils.spark import SharedSparkSession, SparkJob
 
 # Default values for jobs, used per job if not explicitly set in the job's
@@ -132,20 +132,20 @@ def main():
         job.repartition()
 
     # Upload extracted files to AWS S3 in Hive-partitioned Parquet
-    aws = SharedAWSClient(logger=session.logger)
+    aws = AWSClient(logger=session.logger)
     new_local_files = []
     for job in jobs:
-        job_upload_results = job.upload()
+        job_upload_results = job.upload(aws)
         new_local_files.append(job_upload_results)
 
     # If any jobs uploaded never-seen-before files, trigger a Glue crawler
     if any(new_local_files):
         session.logger.info("New partitions uploaded, triggering Glue crawler")
-        aws.run_and_wait_for_crawler("iasworld")
+        aws.run_and_wait_for_crawler("ccao-data-warehouse-iasworld-crawler")
 
     # Trigger a GitHub workflow to run dbt tests once all jobs are complete
     session.logger.info("All file uploads complete, triggering dbt tests")
-    github = SharedGitHubSession(gh_pem_path=PATH_GH_PEM)
+    github = GitHubClient(gh_pem_path=PATH_GH_PEM)
     github.run_workflow(
         repository="https://api.github.com/repos/ccao-data/data-architecture",
         workflow="test_dbt_models.yaml",
--- a/src/submit_jobs.py
+++ b/src/submit_jobs.py
@@ -170,11 +170,10 @@ def submit_jobs(
         workflow="test_dbt_models.yaml",
     )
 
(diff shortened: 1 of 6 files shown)


src/submit_jobs.py
```diff
@@ -123,17 +132,36 @@ def read_job(job: SparkJob):
         job.repartition()
 
     # Upload extracted files to AWS S3 in Hive-partitioned Parquet
+    aws = AWSClient(logger=session.logger)
+    new_local_files = []
```
dfsnow, line 151: This part is a little wacky and I don't love it. Basically, we want to check if any of the jobs found new _local_ files during upload to S3. If they did, then we want to trigger a Glue crawler to pick up the new files/partitions. Open to a better design here though this does work.

src/submit_jobs.py
```diff
@@ -123,17 +132,36 @@ def read_job(job: SparkJob):
...
+
+    # If any jobs uploaded never-seen-before files, trigger a Glue crawler
+    if any(new_local_files):
+        session.logger.info("New partitions uploaded, triggering Glue crawler")
+        aws.run_and_wait_for_crawler("ccao-data-warehouse-iasworld-crawler")
 
     # Trigger a GitHub workflow to run dbt tests once all jobs are complete
-    github = SharedGitHubSession(gh_pem_path=PATH_GH_PEM)
+    session.logger.info("All file uploads complete, triggering dbt tests")
+    github = GitHubClient(logger=session.logger, gh_pem_path=PATH_GH_PEM)
```
dfsnow: Having to pass the logger into the client classes to use it feels kinda of jank. Usually I'd just use `logger = logging.getLogger(__name__)` at the top of each module, but that won't capture the Spark output 😢

src/utils/github.py
```diff
@@ -61,20 +65,28 @@ def create_headers(bearer: str) -> dict:
...
-            )
-            gh_token = response.json()["token"]
-
-            data = {"ref": "master"}
-            response = requests.post(
-                f"{repository}/actions/workflows/{workflow}/dispatches",
-                headers=create_headers(gh_token),
-                json=data,
-            )
+            try:
```
dfsnow, line 68: This stuff is all pretty much the same, just with some exception handling.

src/utils/spark.py
```diff
@@ -231,15 +279,29 @@ def repartition(self) -> None:
...
         )
 
-    def upload(self) -> None:
+        time_end = time.time()
+        time_duration = str(timedelta(seconds=(time_end - time_start)))
+        self.session.logger.info(
+            f"Table {self.table_name} repartitioned in {time_duration}"
+        )
+
+    def upload(self, aws: AWSClient) -> bool:
```
dfsnow: The S3 client setup was originally in here, but it felt weird having a dedicated AWS class AND the client setup here, so I moved the connection the class.

src/utils/spark.py
```diff
@@ -64,12 +65,37 @@ def __init__(self, app_name: str, password_file_path: str) -> None:
...
-        self.s3_client = boto3.client("s3")
-        self.s3_bucket = os.getenv("AWS_S3_BUCKET")
-        self.s3_prefix = os.getenv("AWS_S3_PREFIX", "iasworld")
-
+        # Create the Spark session and logging
         self.spark = SparkSession.builder.appName(self.app_name).getOrCreate()
+        self.log_file_path = f"/tmp/logs/{self.app_name}.log"
+        self.logger = self.get_logger()
+
+    def get_logger(self):
```
dfsnow: This horror is extracting the log4j logger from the JVM and accessing it using a py4j wrapper. This lets us intersperse the Python logger output with the Spark log output from stdout. It also sets up a file sink to write to so that we have a file to upload to CloudWatch.

src/utils/spark.py
```diff
@@ -64,12 +65,37 @@ def __init__(self, app_name: str, password_file_path: str) -> None:
...
+        layout = log4j.PatternLayout()
+        layout.setConversionPattern(
+            "%d{yyyy-MM-dd_HH:mm:ss.SSS} %p %c{1}: %m%n"
+        )
+        appender.setLayout(layout)
+        appender.setFile(self.log_file_path)
+        appender.activateOptions()
+
+        spark_logger.removeAllAppenders()
+        spark_logger.addAppender(appender)
```
dfsnow, line 101: One major annoyance here is that while this setup successfully sinks both Spark and Python log output to the file, only the Spark output is shown in stdout. I couldn't for the life of me figure out how to print both to both file _and_ stdout. That said, I don't think it matters too much since this will be a mostly automatic process and all the logs go to CloudWatch anyways.

src/submit_jobs.py
```diff
@@ -1,9 +1,11 @@
 import argparse
-from datetime import datetime
+import time
+from datetime import datetime, timedelta
 
 from joblib import Parallel, delayed
+from utils.aws import AWSClient
 from utils.helpers import load_job_definitions, load_predicates
-from utils.github import SharedGitHubSession
+from utils.github import GitHubClient
```
jeancochrane, line 7: **[Praise]** Aligning the GitHub class with the `Client` interface is a good call 👍🏻

src/submit_jobs.py
jeancochrane: **[Praise]** I honestly don't mind it! It's a little extra work when constructing clients, but it does make the logger hierarchy clear (i.e. that other services are hooking into the Spark logger).

src/submit_jobs.py
```diff
@@ -123,17 +132,41 @@ def read_job(job: SparkJob):
...
     # Upload extracted files to AWS S3 in Hive-partitioned Parquet
+    aws = AWSClient(logger=session.logger)
+    new_local_files = []
     for job in jobs:
-        job.upload()
+        job_upload_results = job.upload(aws)
+        new_local_files.append(job_upload_results)
+
+    # If any jobs uploaded never-seen-before files, trigger a Glue crawler
+    if any(new_local_files):
```
jeancochrane: **[Suggestion, non-blocking]** I'm a little bit confused by the use of the `SparkJob.upload()` interface here. The fact that we're appending to `new_local_files` makes me think that `SparkJob.upload()` will be returning a string indicating a filename, but instead it returns a boolean. Not super high priority, but here are two alternatives that would be clearer to me:

1. Refactor `SparkJob.upload()` to return a list of strings representing the new local files that were uploaded, then adjust this block accordingly:

```suggestion
    new_local_files = []
    for job in jobs:
        new_local_files_from_job = job.upload(aws)
        if new_local_files_from_job:
            new_local_files += new_local_files_from_job

    # If any jobs uploaded never-seen-before files, trigger a Glue crawler
    if any(new_local_files):
```

2. Keep the `SparkJob.upload()` interface the same, but adjust this block to make it clear that `upload()` returns a bool:

```suggestion
    new_local_files = False
    for job in jobs:
        job_found_new_local_files = job.upload(aws)
        if job_found_new_local_files:
            new_local_files = True

    # If any jobs uploaded never-seen-before files, trigger a Glue crawler
    if new_local_files:
```

Both options work; I would prefer 1 if we think that it might be useful here or elsewhere to know the names of the new local files, and 2 otherwise.

src/submit_jobs.py
```diff
@@ -123,17 +132,41 @@ def read_job(job: SparkJob):
...
-    return session_name
+    # Print table names for extracted tables and total elapsed time. The
+    # get_description() call is to print the description of each table to logs
+    session.logger.info(f"Extracted tables: {', '.join(table_names)}")
+    for job in jobs:
+        job.get_description()
+
+    time_end = time.time()
+    time_duration = str(timedelta(seconds=(time_end - time_start)))
+    session.logger.info(f"Total extraction duration was {time_duration}")
```
jeancochrane: **[Nitpick, non-blocking]** Can we add a unit to the time duration, for clarity (e.g. `s` or `ms`)?

src/utils/aws.py
```diff
@@ -0,0 +1,114 @@
...
+                break
+            elif state == "RUNNING":
+                self.logger.info(
+                    (
+                        f"Crawler {crawler_name} is running: "
+                        f"{time_elapsed}s elapsed"
+                    )
+                )
+            time.sleep(time_increment)
+            time_elapsed += time_increment
```
jeancochrane: **[Suggestion, non-blocking]** I wonder if it's worth adding a generous timeout to this polling block, like 20 mins, just to avoid the possibility that the job hangs overnight if the crawler state transitions in an unexpected way? We could either raise an error if the job times out, or just print a warning:

```suggestion
        # Wait for the crawler to complete before triggering dbt tests
        time_elapsed = 0
        time_increment = 30
        timeout = 1200  # 20 minute timeout
        job_complete = False
        while time_elapsed < timeout:
            response = self.glue_client.get_crawler(Name=crawler_name)
            state = response["Crawler"]["State"]  # type: ignore
            if state in ["READY", "STOPPING"]:
                self.logger.info(f"Crawler {crawler_name} has finished")
                job_complete = True
                break
            elif state == "RUNNING":
                self.logger.info(
                    (
                        f"Crawler {crawler_name} is running: "
                        f"{time_elapsed}s elapsed"
                    )
                )
            time.sleep(time_increment)
            time_elapsed += time_increment
        if not job_complete:
            self.logger.warn(
                f"Crawler {crawler_name} was still running after the {timeout}s polling timeout; "
                "continuing execution without confirming its status"
            )
```

src/utils/github.py
```diff
@@ -61,20 +65,28 @@ def create_headers(bearer: str) -> dict:
...
+                response = requests.post(
+                    f"{repository}/actions/workflows/{workflow}/dispatches",
+                    headers=create_headers(gh_token),
+                    json=data,
+                )
+                response.raise_for_status()
+                self.logger.info(f"GH workflow triggered: {workflow}")
+
+            except Exception as e:
+                self.logger.error(f"GH workflow run failed: {e}")
```
jeancochrane: **[Nitpick, non-blocking]** Is there a good reason to use an acronym here? I think using the full service name will make this log line more search-friendly:

```suggestion
                self.logger.error(f"GitHub workflow run failed: {e}")
```

src/utils/spark.py
jeancochrane: **[Praise]** It's annoying that we have to dig into the PySpark internals to get access to this logger, but I think the end result of a logger that can be passed to other parts of the code to allow them to write to the Spark logs is a solid abstraction. Nice work!

src/utils/spark.py
jeancochrane: **[Praise]** Much cleaner separation of concerns!

jeancochrane (approved):
This rocks! I like the logging interface that you settled on, it feels nice and intuitive. No major concerns from me.

One question I had related to error handling while I was reading this code: A few of the client methods are wrapped in `try`/`except` blocks that prevent them from raising an exception, but what happens if a code block that is not protected in this way causes the job to error out before it can upload its logs to CloudWatch? How will we get alerted that the job has failed? I wonder if it's worth wrapping `main()` in a big `try`/`except` block that logs the exception, attempts to ship logs to CloudWatch, and alerts us before raising the exception.

src/submit_jobs.py
dfsnow: The `timedelta` function actually returns a formatted string like `OH1M14S.100`, so I think we're covered there.

src/submit_jobs.py
dfsnow: Good thinking! I went with 1 and also added a log line about the number of new files uploaded. See afdf098.

dfsnow:
> One question I had related to error handling while I was reading this code: A few of the client methods are wrapped in `try`/`except` blocks that prevent them from raising an exception, but what happens if a code block that is not protected in this way causes the job to error out before it can upload its logs to CloudWatch? How will we get alerted that the job has failed? I wonder if it's worth wrapping `main()` in a big `try`/`except` block that logs the exception, attempts to ship logs to CloudWatch, and alerts us before raising the exception.

So the way I'd set this up it wasn't possible to use the Spark logger in the `except`, since it wouldn't exist if the main loop failed.

In the process of refactoring I discovered you totally can just use both the Spark AND Python loggers at the same time. So, I switched over to generic Python logging in lieu of passing around the Spark logger. This gets us:

- Better namespacing in the log output
- No need to pass around the Spark session logger
- Writing Python logging to both a file and stdout
- An easier to extend logging setup if we add more modules

I think it's a much better design overall, but curious to see what you think. Lots of changes here, so re-requesting review @jeancochrane!

src/utils/helpers.py
```diff
@@ -1,7 +1,53 @@
 import json
+import logging
 from pathlib import Path
 
 
+def create_python_logger(
+    name: str, log_file_path: str = "/tmp/logs/spark.log"
```
dfsnow: The only part I'm unsure about now is that the log file path is specified in like 3 different places: here, the Spark logger, and the CloudWatch upload. I couldn't think of a way to share the file path across all the modules nicely. Any thoughts?

src/utils/spark.py
```diff
@@ -128,6 +156,8 @@ def get_description(self) -> str:
         if self.cur:
             desc.append(f"cur=[{', '.join(self.cur)}]")
 
+        logger.info(f"Table {self.table_name} description: {', '.join(desc)}")
```
jeancochrane: **[Nitpick, non-blocking]** Sorry for not catching this on my first review! I'd recommend that we delegate responsibility for logging the description to the calling code in the `submit_jobs` function, which is already a function with a lot of side effects. IMO a getter method like this one should be as functional as possible, and its callers should not depend on its side effects.

src/utils/spark.py
jeancochrane: **[Question, non-blocking]** Now that we're using a Python logger for the non-Spark code, does the script write Python logs to stdout? Or is it still only Spark logs? I don't think it matters either way, I'm just curious!

src/utils/helpers.py
jeancochrane: **[Suggestion, non-blocking]** I think it's totally fine! As I see it, we're basically just treating `/tmp/logs/spark.log` as the common interface for logs across services, so I think it makes sense that we would reference it in each place where we're making use of that interface. That being said, we could centralize things a tiny bit more if we defined a constant like `SPARK_LOG_FILE = "/tmp/logs/spark.log"` in this file and then imported it into the `utils.spark` and `submit_jobs` modules -- that way if we ever want to change the logfile location we only have to do it once.

jeancochrane (approved):
Awesome work! I think the `create_python_logger` and `create_spark_logger` interfaces make a lot of sense and do a good job of simplifying things.

src/utils/spark.py
dfsnow: Agreed! I moved the logging for the getters to a dedicated block in d1054f6.

src/utils/spark.py
dfsnow: It now writes both Python and Spark logs to both stdout and the log file!

src/utils/helpers.py
dfsnow: Done in 316ddea!

PR #7, opened 2024-08-16
Add CLI args for features

dfsnow:
Quick PR to add command line flags to toggle certain features of the pipeline. This is useful for testing or for building local-only extracts.

### DIFF
--- a/src/submit_jobs.py
+++ b/src/submit_jobs.py
@@ -54,6 +54,40 @@ def parse_args():
         nargs="?",
         help="JSON string containing job configurations(s)",
     )
+    parser.add_argument(
+        "--run-github-workflow",
+        action=argparse.BooleanOptionalAction,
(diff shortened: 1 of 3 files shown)


src/submit_jobs.py
```diff
@@ -187,13 +223,22 @@ def read_job(job: SparkJob):
     app_name = f"iasworld_{current_datetime}"
     args = parse_args()
```
jeancochrane, line 224: **[Suggestion, non-blocking]** What do you think about adding a quick log line here to record all of the values set on `args`?

jeancochrane (approved):
Nice, I agree that these seem helpful for local dev!

src/submit_jobs.py
dfsnow: Done in 612f03e!

PR #8, opened 2024-08-20
Add system for coercing column-level data types

dfsnow:
This PR adds a cascading system for coercing iasWorld data types during Spark extraction. This is necessary in order to match existing data types from Sqoop, create predictable types in Athena, and occasionally coerce the errant column to its proper type.

The PR adds three-levels of schema override, with more specific levels overriding more general ones:

1. Convert all Oracle `NUMERIC` types to `DECIMAL(10,0)` and all `TIMESTAMP` types to `STRING`.[1]
2. Add a global schema override that applies to all tables, defined in `config/default_settings.yaml`.
3. Add table-specific, column-specific overrides, defined in `config/table_definitions.yaml`.

Using this system, I successfully matched all types across the current `iasworld` schema and a temporary `test_iasworld` schema in Athena. I compared types using the script below.

[1]: All Oracle `NUMERIC` columns are read as `DECIMAL(38, 10)` by default. However, AFAICT nearly all such columns are actually integers, and the `NUMERIC` type is simply used to hold them in an unbounded way. Sqoop converts these columns to `DECIMAL(10, 0)` by first reading their value and then assigning the type. There is a non-zero chance that by coercing numbers to this type, we could begin silently truncating values _if_ they ever expand beyond 10,0.

<details>
<summary><strong>R script used to compare data types</strong></summary>

```r
library(noctua)
library(dplyr)
library(glue)
library(stringr)

AWS_ATHENA_CONN_NOCTUA <- dbConnect(noctua::Athena())
tables <- dbGetTables(AWS_ATHENA_CONN_NOCTUA, schema = "iasworld")$TableName

# Extract the types from the string DDL statement
get_col_types <- function(sql) {
  str_extract_all(sql, "`[a-z0-9_-]+` [a-z0-9\\)\\(,]*[,)]") %>%
    .[[1]] %>%
    str_remove_all("\\)$") %>%
    str_remove_all(",$") %>%
    str_remove_all("`") %>%
    tibble() %>%
    tidyr::separate(col = ".", into = c("col", "type"), sep = " ")
}

# Grab new and old DDLs for each table and munge them into one dataframe
types_df <- purrr::map_dfr(tables, \(table) {
  print(paste("scanning table:", table))
  old_ddl <- dbShow(AWS_ATHENA_CONN_NOCTUA, glue("iasworld.{table}"))
  new_ddl <- dbShow(AWS_ATHENA_CONN_NOCTUA, glue("test_iasworld.{table}"))

  old <- get_col_types(old_ddl)
  new <- get_col_types(new_ddl)

  merged <- old %>%
    left_join(new, by = "col", suffix = c("_old", "_new")) %>%
    mutate(table = {{ table }}) %>%
    relocate(table, .before = everything()) %>%
    mutate(no_match = type_old != type_new)

  merged
})
```
</details>

### DIFF
--- a/config/default_settings.yaml
+++ b/config/default_settings.yaml
@@ -13,5 +13,5 @@ default_args:
 #   2. Global schema overrides, as defined below
 #   3. Table schema overrides, defined in table_definitions.yaml
 global_schema_overrides:
-  - deactivat: STRING
-  - taxyr: STRING
+  deactivat: STRING
+  taxyr: STRING
--- a/config/default_settings.yaml
+++ b/config/default_settings.yaml
@@ -1,10 +1,10 @@
 # Default arguments used for the CLI. Can be overridden using the corresponding
 # flag at runtime
 default_args:
-  run_github_workflow: False
-  run_glue_crawler: False
-  upload_logs: False
-  upload_data: False
+  run_github_workflow: True
+  run_glue_crawler: True
+  upload_logs: True
+  upload_data: True
 
 # Data type overrides are necessary to maintain parity/consistency with sqoop.
 # Schema overrides are defined in three places, with each overriding the last:
--- a/config/default_settings.yaml
+++ b/config/default_settings.yaml
@@ -1,14 +1,16 @@
 # Default arguments used for the CLI. Can be overridden using the corresponding
 # flag at runtime
 default_args:
-  run-github-workflow: True
-  run-glue-crawler: True
-  upload-logs: True
-  upload-data: True
+  run_github_workflow: False
+  run_glue_crawler: False
+  upload_logs: False
+  upload_data: False
 
-# Schema coercions used for every table. Note that these are used
-# even if the column does not exist in a given table; Spark will simply
-# ignore such overrides
+# Data type overrides are necessary to maintain parity/consistency with sqoop.
+# Schema overrides are defined in three places, with each overriding the last:
+#   1. By default, NUMBER Oracle types are converted to DECIMAL(10,0)
+#      and TIMESTAMP Oracle types are converted to STRING
+#   2. Global schema overrides, as defined below
+#   3. Table schema overrides, defined in table_definitions.yaml
 global_schema_overrides:
-  - iasw_id DECIMAL(38, 0)
-  - wen STRING
+  - deactivat: STRING
--- a/config/default_settings.yaml
+++ b/config/default_settings.yaml
@@ -14,3 +14,4 @@ default_args:
 #   3. Table schema overrides, defined in table_definitions.yaml
 global_schema_overrides:
   - deactivat: STRING
+  - taxyr: STRING
--- a/config/default_settings.yaml
+++ b/config/default_settings.yaml
@@ -1,0 +1,14 @@
+# Default arguments used for the CLI. Can be overridden using the corresponding
+# flag at runtime
+default_args:
+  run-github-workflow: True
+  run-glue-crawler: True
+  upload-logs: True
+  upload-data: True
+
+# Schema coercions used for every table. Note that these are used
+# even if the column does not exist in a given table; Spark will simply
+# ignore such overrides
+global_schema_overrides:
+  - iasw_id DECIMAL(38, 0)
+  - wen STRING
--- a/config/table_definitions.yaml
+++ b/config/table_definitions.yaml
@@ -7,136 +7,170 @@ tables:
     min_year: null
     has_cur: false
     has_parid: false
-    schema_overrides:
-      - ss_ext_date STRING
+
   addn:
     min_year: *global_min_year
     has_cur: true
     has_parid: true
+
   addrindx:
     min_year: *global_min_year
     has_cur: true
     has_parid: true
+    schema_overrides:
+      - xcoord: DECIMAL(15,8)
+      - ycoord: DECIMAL(15,8)
+
   aprval:
     min_year: *global_min_year
     has_cur: true
     has_parid: true
+
   asmt_all:
     min_year: *global_min_year
     has_cur: true
     has_parid: true
+
   asmt_hist:
     min_year: *global_min_year
     has_cur: true
     has_parid: true
+
   cname:
     min_year: null
     has_cur: true
     has_parid: false
+
   comdat:
     min_year: *global_min_year
     has_cur: true
     has_parid: true
+
   comnt:
     min_year: null
     has_cur: false
     has_parid: false
+
   cvleg:
     min_year: 1900
     has_cur: false
     has_parid: true
+
   cvown:
     min_year: 2001
     has_cur: false
     has_parid: false
+
   cvtran:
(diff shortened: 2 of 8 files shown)


config/default_settings.yaml
```diff
@@ -0,0 +1,17 @@
+# Default arguments used for the CLI. Can be overridden using the corresponding
+# flag at runtime
+default_args:
+  run_github_workflow: True
+  run_glue_crawler: True
+  upload_logs: True
+  upload_data: True
```
dfsnow, line 7: I defined these here because it was nice to be able to quickly change runtime flags here instead of in the CLI. Open to removing them if this is too confusing/weird!

config/table_definitions.yaml
dfsnow, line 1: This new file defines two things for each table:

1. The possible parameters for each table e.g. whether or not the table has a `cur` column that can be used in a job definition. These will be used for a simple job validator in a future PR.
2. Table-level schema overrides, used to coerce columns to their expected types.

config/table_definitions.yaml
```diff
@@ -0,0 +1,155 @@
...
+# as well as table-level schema overrides per column. Null values indicate
+# the column is not applicable to the table, or that there should be no
+# check on the possible value
+tables:
+  aasysjur:
+    min_year: null
+    has_cur: false
+    has_parid: false
+  addn:
+    min_year: *global_min_year
```
dfsnow, line 14: I plan to use a non-null `min_year` value here implicitly to check if tables have a `taxyr` column, but will save that for another PR.

src/utils/helpers.py
```diff
@@ -2,9 +2,67 @@
...
+
+
+def convert_decimal_columns(
+    df: DataFrame, ignore_cols: list[str]
+) -> DataFrame:
+    """
+    iasWorld stores many columns as the generic NUMBER Oracle data type, but
+    almost always uses DECIMAL(10,0) for the actual values. This function
+    automatically converts all such columns to DECIMAL(10,0). Columns specified
+    via schema overrides will ignore this conversion.
```
dfsnow: Note my warning in the PR body: there's a chance that by doing this we could end up silently truncating numbers in some future where they expand beyond `DECIMAL(10,0)`. 

We _could_ simply leave everything as `DECIMAL(38,10)` to avoid this problem, but the downside is that Athena will then show/return numbers like `1242424.0000000000` even for small integers.

I do not understand why they designed their database this way 🙃

config/default_settings.yaml
jeancochrane: **[Praise]** I like this! In my experience, it's a common pattern for CLIs to allow for overriding values using a config file.

config/default_settings.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+# Data type overrides are necessary to maintain parity/consistency with sqoop.
+# Schema overrides are defined in three places, with each overriding the last:
+#   1. By default, NUMBER Oracle types are converted to DECIMAL(10,0)
+#      and TIMESTAMP Oracle types are converted to STRING
+#   2. Global schema overrides, as defined below
+#   3. Table schema overrides, defined in table_definitions.yaml
+global_schema_overrides:
+  - deactivat: STRING
+  - taxyr: STRING
```
jeancochrane: **[Suggestion, non-blocking]** Unless we imagine that it would be useful to allow a user to configure multiple overrides for one field, it might be cleaner to format this attribute as a key-value mapping rather than a list of key-value mappings:

```suggestion
global_schema_overrides:
  deactivat: STRING
  taxyr: STRING
```

src/utils/helpers.py
```diff
@@ -2,9 +2,67 @@
...
+    Returns:
+        Spark DataFrame with TIMESTAMP columns converted to STRING.
+    """
+    for field in df.schema.fields:
+        if (
+            isinstance(field.dataType, TimestampType)
+            and field.name not in ignore_cols
+        ):
+            df = df.withColumn(field.name, df[field.name].cast(StringType()))
+    return df
```
jeancochrane: **[Thought, non-blocking]** This might just be a personal preference, but I dislike the pandas convention of defining single-use functions that take in a dataframe, transform a bunch of the dataframe's columns, and then return a new copy of the dataframe with transformed columns. I think the intention behind the pattern is to make the top-level script easier to read by summarizing each transformation and factoring it out into a separate function, but as a reader I tend to find the pattern _more_ confusing than the alternative because the lack of references to column names or types means that the function interface doesn't actually tell you what columns are undergoing transformation or what those transformations should look like.

Still, I recognize that this is a common pandas pattern, and I'm a lot newer to writing production pandas than you are! I'm just curious if you've heard this critique before and if you have thoughts about the right approach to readability for complex pandas transformations. My impression is that R solves this problem with the tidyverse, which provides a bunch of nice abstractions around data transformations that make it unnecessary to factor logic out into isolated functions in the majority of cases. I'm not sure if there's an equivalent approach to pandas code but I'd be interested in learning more about it.

config/table_definitions.yaml
jeancochrane: **[Suggestion, non-blocking]** Nice! Should we add these two new config files to the README file structure diagram?

config/default_settings.yaml
```diff
@@ -0,0 +1,17 @@
...
+  run_glue_crawler: True
+  upload_logs: True
+  upload_data: True
+
+# Data type overrides are necessary to maintain parity/consistency with sqoop.
+# Schema overrides are defined in three places, with each overriding the last:
+#   1. By default, NUMBER Oracle types are converted to DECIMAL(10,0)
+#      and TIMESTAMP Oracle types are converted to STRING
+#   2. Global schema overrides, as defined below
+#   3. Table schema overrides, defined in table_definitions.yaml
```
jeancochrane, line 14: **[Suggestion, non-blocking]** Very nice! At some point it might be nice to document this config system in the top-level README, but I don't feel strongly about that happening right now. We might just want to take a final pass at the docs once this repo is in a stable state.

jeancochrane (approved):
Awesome job! I like the simplicity of the config structure, and the inheritance makes sense to me.

config/table_definitions.yaml
dfsnow: Will do! I have a follow-up PR to update the README once this is merged.

config/default_settings.yaml
dfsnow: I have a follow-up PR for the README right after this one!

config/default_settings.yaml
dfsnow: Good thinking! Also lets us drop a helper function. Fixed in 770c37b.

src/utils/helpers.py
dfsnow: I also dislike it and would usually in-line code like this, but in this case I wanted to call out and document what's happening since it's pretty important/weird behavior. I think instead I'll move these inline and add a note to the README then link to it.

PR #10, opened 2024-08-21
Add fast follows for first full run

dfsnow:
This PR bundles a few small fixes resulting from our first full Spark extraction.


src/submit_jobs.py
```diff
@@ -20,7 +20,7 @@
 # Default values for jobs, used per job if not explicitly set in the job's
 # input JSON. CUR and YEAR values are used for partitioning and filtering
 # if they are set
-DEFAULT_VAR_MIN_YEAR = 1999
+DEFAULT_VAR_MIN_YEAR = 1800
```
dfsnow, line 23: There are tables with `taxyr` as early as 1900, so we want the unbounded case (where `min_year` is NULL) to capture those earliest records.

src/utils/aws.py
```diff
@@ -99,6 +99,9 @@ def upload_logs_to_cloudwatch(
                         }
                     )
 
+            # Sort log events by timestamp
+            log_events.sort(key=lambda event: event["timestamp"])
```
dfsnow, line 103: With both Spark and Python writing to the same log file, we sometimes get timestamps that are slightly out of order. When that happens, CloudWatch throws an error on upload. This orders all log records to prevent that.

src/utils/spark.py
```diff
@@ -173,7 +174,8 @@ def get_filter(self) -> str | None:
         """
         filter = []
         if self.taxyr:
-            filter.append(f"taxyr IN ({', '.join(map(str, self.taxyr))})")
+            min_year, max_year = self.taxyr[0], self.taxyr[-1]
+            filter.append(f"taxyr BETWEEN {min_year} AND {max_year}")
```
dfsnow, line 178: Using BETWEEN instead of IN makes for cleaner logs, is probably faster, and yields the same query results.

src/utils/spark.py
```diff
@@ -261,6 +263,12 @@ def read(self) -> None:
                     field.name, df[field.name].cast(DecimalType(10, 0))
                 )
 
+        # Add a timestamp column to the data to track when it was loaded
+        df = df.withColumn(
+            "loaded_at",
+            date_format(current_timestamp(), "yyyy-MM-dd HH:mm:ss.SSS"),
+        )
+
```
dfsnow, line 271: I somehow dropped the `loaded_at` column in an earlier PR, but we need it for dbt source freshness tests and for parity with sqoop.

jeancochrane (approved):
Looks great!

PR #13, opened 2024-08-30
Enable extracting iasWorld test environment

dfsnow:
This PR modifies the Spark extractor to allow extracting from the iasWorld test environment. Previously, we only extracted from the production environment.

This setup is necessary to facilitate https://github.com/ccao-data/enterprise-intelligence/issues/131, which requires pulling data that is only loaded into the test environment.

Data pulled from the test environment is put into its own dedicated database in Athena, `iasworld_test`. Note that the test environment doesn't replicate the production environment; not all tables are available in test, and the most recent year of data is not present.

I've already successfully run this pipeline, so you can find the test data in Athena under `iasworld_test`.


README.md
```diff
@@ -115,6 +115,8 @@ docker exec spark-node-master ./submit.sh --json-file /tmp/jobs.json
 
 The command line interface also has multiple optional flags:
 
+- `--extract-target` - iasWorld target environment to extract data from. Must
```
dfsnow, line 118: I don't expect that we'll need to pull data from any further environments (unless there are some I don't know about), so keeping this limited to two choices.

README.md
```diff
@@ -207,6 +209,9 @@ user 1003.
 
 # Extract all tables on Saturday at 1 AM CST
 0 6 * * 6 docker exec spark-node-master ./submit.sh --json-string "$(yq -o=json .weekend_jobs /full/path/to/default_jobs.yaml)"
+
+# Extract all test environment tables on Sunday at 1 AM CST
+0 6 * * 7 docker exec spark-node-master ./submit.sh --json-string "$(yq -o=json .weekend_jobs_test /full/path/to/default_jobs.yaml)" --extract-target test
```
dfsnow, line 214: I figure that extracting the test environment once a week is probably enough.

.env
```diff
@@ -2,8 +2,12 @@ AWS_S3_BUCKET=
...
-IPTS_SERVICE_NAME=
-IPTS_USERNAME=
+IPTS_PRD_HOSTNAME=
+IPTS_PRD_PORT=
+IPTS_PRD_SERVICE_NAME=
+IPTS_PRD_USERNAME=
+IPTS_TST_HOSTNAME=
+IPTS_TST_PORT=
+IPTS_TST_SERVICE_NAME=
+IPTS_TST_USERNAME=
```
dfsnow, line 12: We now need two different sets of credentials/login details.

src/submit_jobs.py
```diff
@@ -122,9 +132,12 @@ def submit_jobs(
 
     # Each session is shared across all read jobs and manages job order,
     # credentialing, retries, etc.
+    target = "TST" if extract_target == "test" else "PRD"
```
dfsnow, line 134: I don't _love_ just adding simple conditional logic inline like this, but I also don't expect that we'll need to target environments any time soon. So, IMO it's fine for now.

src/submit_jobs.py
```diff
@@ -240,14 +258,23 @@ def read_job(job: SparkJob):
...
+    app_name = f"iasworld_{args.extract_target}_{current_datetime}"
 
     try:
+        # Clear the existing extract files and logs before submitting new jobs.
+        # This is to prevent new jobs from becoming mixed with the results of
+        # previous failed or cancelled jobs
+        clear_directory(PATH_INITIAL_DIR)
+        clear_directory(PATH_FINAL_DIR)
+        if os.path.exists(PATH_SPARK_LOG):
+            os.remove(PATH_SPARK_LOG)
```
dfsnow: We _really_ don't want to get the test and prod data mixed up, so I'm clearing the extract directory at the start of every run.

src/utils/helpers.py
```diff
@@ -1,12 +1,33 @@
...
 import logging
+import shutil
 from pathlib import Path
 
 import yaml
 
 PATH_SPARK_LOG = "/tmp/logs/spark.log"
 
 
+def clear_directory(dir: Path | str) -> None:
```
dfsnow, line 11: I switched back to this helper function because `shutil.rmtree` also deletes the directory itself, and I wanted to preserve the directory structure and `.gitkeep` file of `target/initial` and `target/final`.

config/default_jobs.yaml
```diff
@@ -189,6 +189,80 @@ weekend_jobs:
     cur: null
     predicates_path: null
 
+# Jobs that run on the weekend targeting the test environment. Note
+# that the subset of tables we have access to in test is more limited
+# than in prod i.e. we only have access to the tables below
+weekend_jobs_test:
+  addn:
+    table_name: ias.addn
```
dfsnow, line 197: Note that the schema is also slightly different here, so glad we made `table_name` a job option.

src/submit_jobs.py
```diff
@@ -1,12 +1,15 @@
 import argparse
+import os
 import time
 from datetime import datetime, timedelta
 
 from joblib import Parallel, delayed
+
```
jeancochrane, line 6: **[Nitpick, non-blocking]** I think this is a stray newline, yeah?

```suggestion
```

src/utils/helpers.py
```diff
@@ -1,14 +1,35 @@
...
+    for item in dir_path.iterdir():
+        if item.is_dir():
+            shutil.rmtree(item.as_posix())
+        elif item.name != ".gitkeep":
+            item.unlink()
+
+
 def create_python_logger(
-    name: str, log_file_path: str = PATH_SPARK_LOG
+    name: str, mode: str = "a", log_file_path: str = PATH_SPARK_LOG
```
dfsnow, line 32: This was necessary in order to correctly handle file truncation via the logging module. The first logger that gets called (at the top of `submit_jobs.py` uses `mode="w"` to truncate the existing file, then all subsequent loggers use `mode="a"`.

src/submit_jobs.py
jeancochrane: **[Praise]** Agreed! I also don't think it'll be too difficult to refactor in the future if this assumption changes.

jeancochrane (approved):
Nice work!

src/utils/helpers.py
jeancochrane: **[Thought, non-blocking]** I wonder if there's a way we can accomplish the same goal without having to rely on the assumption that certain parts of the code will run before others? Not a huge deal, if it works we can always iterate on it later.

dfsnow: In this case, it's not actually too bad if the code happened to run out-of-order, as we're basically only using the write mode to clear the previous file. If stuff ran out-of-order somehow, we'd just end up appending to the file instead of truncating it first.

I tried deleting it manually before writing, but Python gets mad when the file it was writing to ceases to exist _after_ the logger has started.

src/submit_jobs.py
dfsnow: Ruff put that in there I think. There's supposed to be a separating between third-party and first-party module imports.

jeancochrane: My bad, that was obvious and I have no idea why I missed it 🤦🏻‍♀️

PR #14, opened 2024-09-02
Fix log file deletion/write mode

dfsnow:
We had some Spark alarms this weekend indicating that our jobs didn't run. They actually did, but the logs weren't written to CloudWatch due to changes from 2212877054a983ad484d735b01aeae1163cc8cee. This PR reverts those changes and moves the log removal to the beginning of the job removal script.

### DIFF
--- a/src/submit_jobs.py
+++ b/src/submit_jobs.py
@@ -1,4 +1,5 @@
 import argparse
+import os
 import time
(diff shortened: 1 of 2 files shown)


jeancochrane (approved):
Thanks for debugging that! This approach makes more intuitive sense to me anyway.

PR #16, opened 2024-09-06
Fix filter pushdown

dfsnow:
This PR fixes filter pushdown for tables using `taxyr`. Previously, forcing the `taxyr` schema value to STRING caused filtering to be done _after_ this initial SQL pull.


jeancochrane (approved):
Brilliant catch!

PR #18, opened 2024-09-24
Set var to upload test results when running GitHub workflow

jeancochrane:
**⚠️ This PR depends on changes in https://github.com/ccao-data/data-architecture/pull/603. Make sure to merge that PR before this one. ⚠️** 

This PR updates the GitHub workflows integration in our Spark ingest config to pass the `upload_test_results=True` input variable when triggering the `test-dbt-models` workflow in https://github.com/ccao-data/data-architecture/. This parameter will be necessary in order to run the workflow properly once https://github.com/ccao-data/data-architecture/pull/603 lands.

Is there an easy way to test this change other than pulling this branch on the server and running a manual test? Perhaps I could clone a new copy of the repo, copy over the secrets, build the containers, and then run a test from the test repo? Let me know what you think the best path forward is for testing and I'll make sure to do it before pulling this PR in.

### DIFF
--- a/src/submit_jobs.py
+++ b/src/submit_jobs.py
@@ -248,6 +248,7 @@ def submit_jobs(
         github.run_workflow(
             repository=(github.gh_api_url + "ccao-data/data-architecture"),
             workflow="test_dbt_models.yaml",
+            inputs={"upload_test_results": True},
         )
 
     # Print table names and descriptions for extracted tables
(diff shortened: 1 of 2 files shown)


dfsnow (approved):
Code changes here look good. Re: testing, let's just recursively copy the repo on the server (secrets and all) to a new directory, switch branches, test, then delete it. Not ideal, but fine for now.

PR #20, opened 2024-11-07
Update `github.run_workflow()` call to specify which tests to run

jeancochrane:
**⚠️ This PR depends on changes in https://github.com/ccao-data/data-architecture/pull/638. Make sure to merge that PR before this one. ⚠️** 

This companion PR of https://github.com/ccao-data/data-architecture/pull/638/ implements a simple one-line change to specify the [dbt selector](https://docs.getdbt.com/reference/node-selection/yaml-selectors) representing the tests that the Spark process should run when the `--run-github-workflow` option is enabled.

This change is necessary because https://github.com/ccao-data/data-architecture/pull/638/ changes the default set of tests that the workflow will run in the absence of a `select` or `selector` option.

### DIFF
--- a/src/submit_jobs.py
+++ b/src/submit_jobs.py
@@ -248,7 +248,10 @@ def submit_jobs(
         github.run_workflow(
             repository=(github.gh_api_url + "ccao-data/data-architecture"),
             workflow="test_dbt_models.yaml",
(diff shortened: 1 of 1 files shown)


dfsnow (approved):
Nice 👍

PR #22, opened 2024-11-13
Setup Spark development environment

### LINKED ISSUE #21: Create Spark dev environment
It would also be nice to have an easy way to spin up a development environment to manually test changes. Currently there are a lot of configs that are only set in the deployed copy of the repo, and the prod config can't easily be copied and reused without editing a bunch of files to avoid conflicts with the running prod instance.

For future reference, here are the changes I had to make to get a test environment up and running after copying over the repo:

```diff
diff --git a/config/spark-defaults.conf b/config/spark-defaults.conf
index 5f1e64f..2286b53 100644
--- a/config/spark-defaults.conf
+++ b/config/spark-defaults.conf
@@ -4,7 +4,7 @@ spark.driver.extraClassPath  /jdbc/ojdbc8.jar
 spark.driver.extraClassPath  /jdbc/ojdbc8.jar

 # Docker hostname address of the Spark master node
-spark.master          spark://spark-node-master:7077
+spark.master          spark://spark-node-master-test:7077
 # Driver memory can be low, as it's only used for the JDBC connection
 spark.driver.memory   2g
 # Total memory available to the worker, across all jobs
diff --git a/docker-compose.yaml b/docker-compose.yaml
index 4f72205..b0ce44c 100644
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -2,11 +2,10 @@ x-default: &node
   image: ${CCAO_REGISTRY_URL}/service-spark-iasworld:latest
   # Corresponds to the user set in the Dockerfile and the shiny-server
   # user on the server for proper write perms to mounted directories
-  user: 1003:0
+  user: 1010:0
   build:
     context: .
     dockerfile: Dockerfile
-  restart: unless-stopped
   volumes:
     - ./drivers:/jdbc:ro
     - ./src:/tmp/src:ro
@@ -20,13 +19,13 @@ x-default: &node
     - IPTS_PRD_PASSWORD
     - IPTS_TST_PASSWORD
   networks:
-    - sparknet
+    - sparknet-test

 services:
-  spark-node-master:
+  spark-node-master-test:
     <<: *node
-    container_name: spark-node-master
-    hostname: spark-node-master
+    container_name: spark-node-master-test
+    hostname: spark-node-master-test
     environment:
       - SPARK_MODE=master

@@ -49,16 +48,16 @@ services:
       - IPTS_TST_PORT
       - GH_APP_ID
     ports:
-      - 4040:4040
-      - 8080:8080
+      - 4041:4040
+      - 8081:8080

-  spark-node-worker:
+  spark-node-worker-test:
     <<: *node
-    container_name: spark-node-worker
-    hostname: spark-node-worker
+    container_name: spark-node-worker-test
+    hostname: spark-node-worker-test
     environment:
       - SPARK_MODE=worker
-      - SPARK_MASTER_URL=spark://spark-node-master:7077
+      - SPARK_MASTER_URL=spark://spark-node-master-test:7077
       - SPARK_WORKER_MEMORY=96G
       - SPARK_WORKER_CORES=28
       - SPARK_WORKER_WEBUI_PORT=9090
@@ -82,16 +81,16 @@ services:
       - IPTS_TST_PORT
       - GH_APP_ID
     ports:
-      - 9090:9090
+      - 9091:9090

 # Using a dedicated subnet because the Docker default subnet conflicts
 # with some of the CCAO's internal routing
 networks:
-  sparknet:
+  sparknet-test:
     ipam:
       config:
-        - subnet: 211.55.0.0/16
-    name: sparknet
+        - subnet: 211.54.0.0/16
+    name: sparknet-test

 secrets:
   AWS_CREDENTIALS:
diff --git a/src/utils/github.py b/src/utils/github.py
index 600c0e9..5210e8d 100644
--- a/src/utils/github.py
+++ b/src/utils/github.py
@@ -84,7 +84,7 @@ class GitHubClient:
                 response.raise_for_status()
                 gh_token = response.json()["token"]

-                data: dict[str, str | dict] = {"ref": "master"}
+                data: dict[str, str | dict] = {"ref": "jeancochrane/gate-test-result-s3-upload-behind-workflow-variable"}
                 if inputs is not None:
                     data["inputs"] = inputs
```

_Originally posted by @jeancochrane in https://github.com/ccao-data/service-spark-iasworld/issues/12#issuecomment-2463157088_

dfsnow:
This PR uses [Docker Compose profiles](https://docs.docker.com/compose/how-tos/profiles/) to set up a separate development version of the production Spark containers. This will allow users on the same machine (i.e. the server) to spin up a parallel set of containers for Spark development, without interfering/overlapping with the existing environment.

Note that at least for now, the dev environment has the same targets as the production environment. It will upload things to the same S3 bucket/log group and trigger the production crawler/workflows. We can amend this in the future with a separate PR.

There's a _lot_ of fiddly stuff in this PR. I'll try to call it out in line level comments.

Closes #21.


Dockerfile
```diff
@@ -15,7 +15,11 @@ RUN mkdir -p /tmp/python && \
...
-# User corresponds to dedicated shiny-server user on server. Necessary for
-# proper write permissions to target directories
-USER 1003:0
+# User corresponds to dedicated shiny-server user on server OR the user
+# specified via build args. Necessary for proper write permissions to target
+# directories. Note that the group must remain root due to issues with the
+# upstream container
+ARG USER_ID=1003
+
+USER ${USER_ID}:0
```
dfsnow: Whoever is hacking on this repo will need to run `docker compose build` to create a version of the container that users their own UID. This will let them remove/edit files created by Spark in the `target/` directory (which would otherwise be owned by `shiny-server` or `root`).

README.md
```diff
@@ -102,15 +102,15 @@ For example, to submit the test jobs in `config/default_jobs.yaml` via
 `--json-string`, run the following command:
 
 ```bash
-docker exec spark-node-master ./submit.sh \
+docker exec spark-node-master-prod ./submit.sh \
     --json-string "$(yq -o=json .test_jobs ./config/default_jobs.yaml)"
```
dfsnow: Cron jobs calling Spark will now need to specify a `-dev` or `-prod` suffix to target the new container names.

README.md
```diff
@@ -194,60 +194,106 @@ populate the following:
...
 schedule file, use `crontab -e` as the main server user. The example crontab
 file below schedules daily jobs for frequently updated tables and weekly ones
-for rarely-updated tables. Note that the jobs currently _must_ be run as
-user 1003.
+for rarely-updated tables.
 
 ```bash
 # Extract recent years from frequently used tables on weekdays at 1 AM CST
-0 6 * * 1,2,3,4,5 docker exec spark-node-master ./submit.sh --json-string "$(yq -o=json .default_jobs /full/path/to/default_jobs.yaml)"
+0 6 * * 1,2,3,4,5 docker exec spark-node-master-prod ./submit.sh --json-string "$(yq -o=json .default_jobs /full/path/to/default_jobs.yaml)" --run-github-workflow --run-glue-crawler --upload-data --upload-logs
```
dfsnow, line 251: Note that we need to explicitly set all these flags since they're now `False` by default. We'll also need to update the actual cron jobs once this is merged.

README.md
```diff
@@ -194,60 +194,106 @@ populate the following:
...
-0 6 * * 7 docker exec spark-node-master ./submit.sh --json-string "$(yq -o=json .weekend_jobs_test /full/path/to/default_jobs.yaml)" --no-run-github-workflow --extract-target test
+0 6 * * 7 docker exec spark-node-master-prod ./submit.sh --json-string "$(yq -o=json .weekend_jobs_test /full/path/to/default_jobs.yaml)" --no-run-github-workflow --run-glue-crawler --upload-data --upload-logs --extract-target test
 ```
 
 ## Structure
 
 Here's a breakdown of important files and the purpose of each one:
 
-```tree
+```bash
```
dfsnow, line 264: This is just a totally unrelated formatting change I made because I think it's easier to read.

config/spark-defaults.conf
```diff
@@ -1,3 +1,6 @@
+# These are config options shared across ALL Spark nodes and are used
+# when instantiating the node i.e. when running docker compose up -d
+
```
dfsnow, line 3: I tried to call out which configuration options are doing what, but it's _really_ confusing. There are global options set across all nodes (this file), per worker/master options set by env vars (in compose), and per run options set in `src/submit.sh`.

docker-compose.yaml
```diff
@@ -19,79 +19,126 @@ x-default: &node
...
-      - IPTS_PRD_SERVICE_NAME
-      - IPTS_PRD_HOSTNAME
-      - IPTS_PRD_USERNAME
-      - IPTS_PRD_PORT
-      - IPTS_TST_SERVICE_NAME
-      - IPTS_TST_HOSTNAME
-      - IPTS_TST_USERNAME
-      - IPTS_TST_PORT
-      - GH_APP_ID
+      <<: [*spark-env]
```
dfsnow: This bizarre syntax seems specific to Compose and lets you merge maps i.e. combine all the env vars above with the ones specified here.

docker-compose.yaml
```diff
@@ -1,12 +1,12 @@
...
   # Corresponds to the user set in the Dockerfile and the shiny-server
   # user on the server for proper write perms to mounted directories
-  user: 1003:0
+  user: "${UID:-1003}:0"
   build:
     context: .
     dockerfile: Dockerfile
-  restart: unless-stopped
+    args:
+      - USER_ID=${UID:-1003}
```
dfsnow: These UID args pass into the container during build and run time. I'm not positive that it's required to specify the build time user in the Dockerfile, but it can't hurt either.

docker-compose.yaml
```diff
@@ -19,79 +19,126 @@ x-default: &node
...
 
 services:
-  spark-node-master:
+  spark-node-master-prod:
     <<: *node
-    container_name: spark-node-master
+    profiles: ["prod", ""] # empty string here runs when no profile is specified
+    image: ${CCAO_REGISTRY_URL}/service-spark-iasworld:latest
+    container_name: spark-node-master-prod
     hostname: spark-node-master
```
dfsnow: The hostname is the same across both `dev` and `prod` because it's internal to their respective Docker networks.

docker-compose.yaml
```diff
@@ -19,79 +19,126 @@ x-default: &node
...
+      SPARK_WORKER_WEBUI_PORT: 9090
     ports:
       - 9090:9090
+    networks:
+      - sparknet-prod
+
+  spark-node-master-dev:
+    <<: *node
+    profiles: ["dev"]
+    image: ${CCAO_REGISTRY_URL}/service-spark-iasworld:dev
```
dfsnow: Note the different image tag for the dev containers. This will force a `docker compose build` on first run, which ensures the `dev` container reflects the new user and any changes made to the Dockerfile or dependencies.

src/submit_jobs.py
```diff
@@ -60,49 +60,49 @@ def parse_args(defaults) -> argparse.Namespace:
         "--json-file",
         type=str,
         nargs="?",
-        help="""
-        Path to a JSON file containing job configuration(s)". The path is
-        relative to the project `config/` directory
-        """,
```
dfsnow, line 66: Kind of interesting: I'm pretty sure this is auto-formatting from ruff designed to escape the double quote. I definitely wouldn't have used triple quotes here.

docker-compose.yaml
jeancochrane: **[Thought, non-blocking]** Seems like [the compose `user` attribute overrides the env var set in the container](https://docs.docker.com/reference/compose-file/services/#user), so I think we could get away with only setting it in one of the two places, but I agree that I can't see how it would hurt to set both:

> `user` overrides the user used to run the container process. The default is set by the image (i.e. Dockerfile `USER`). If it's not set, then `root`.

docker-compose.yaml
jeancochrane: **[Question, non-blocking]** I'm confused why we need to use the array syntax here, but not in the similar cases where we're merging keys from the `node` alias into top-level service attributes. If I'm reading [the Compose docs](https://docs.docker.com/reference/compose-file/extension/#example-4) and [the merge key working draft](https://yaml.org/type/merge.html) correctly, the array syntax is only necessary when merging multiple aliases at once:

> If the value associated with the key is a single mapping node, each of its key/value pairs is inserted into the current mapping, unless the key already exists in it. If the value associated with the merge key is a sequence, then this sequence is expected to contain mapping nodes and each of these nodes is merged in turn according to its order in the sequence.

So in this case we're merging a "sequence", whereas the pre-existing merges like `<<: *node` represented a merge of a "single mapping node". It doesn't sound to me like there's a functional difference between the way these two operations merge mapping nodes, other than that sequences allow for merging multiple nodes at once, with later nodes in the sequence potentially overriding former nodes. Does that match your understanding? If so, it doesn't really matter if we keep the sequence syntax or not, but I think we should at least clear up the comments in the file to make it clear how this works.

As an aside, [this SO answer](https://stackoverflow.com/a/9525884) provides an interesting (but hypothetical) motivation for the sequence syntax:

> [S]ince `<<` is a key, it can only be specified once per node.

docker-compose.yaml
```diff
@@ -19,79 +19,126 @@ x-default: &node
     - GH_PEM
     - IPTS_PRD_PASSWORD
     - IPTS_TST_PASSWORD
-  networks:
-    - sparknet
+
+# These environmental variables get templated/merged into the env vars of
+# containers. Usually YAML doesn't let you merge arrays, but Compose allows it:
+# https://docs.docker.com/reference/compose-file/extension/#example-4
```
jeancochrane: **[Nitpick, non-blocking]** See my comment below for more details about the merge key syntax, but TL;DR I don't actually think this comment is true -- the merge key syntax does allow you to merge sequences (basically another name for arrays of nodes) but those sequences still need to contain nodes and so we can't actually use them to merge arrays. The [Docker docs](https://docs.docker.com/reference/compose-file/extension/#example-4) seem to say as much in the note:

> [YAML merge](https://yaml.org/type/merge.html) only applies to mappings, and can't be used with sequences.

I also don't think this behavior is specific to Compose, since it's part of the YAML spec. Let me know if you want to discuss further to clear this up, I admit that the way it's documented for both Compose and the YAML spec is super confusing.

docker-compose.yaml
```diff
@@ -19,79 +19,126 @@ x-default: &node
...
+    hostname: spark-node-master
+    restart: no
+    environment:
+      <<: [*spark-env]
+      SPARK_MODE: master
+      SPARK_UI_PORT: 4042
+      SPARK_EXECUTOR_MEMORY: 48g
+      # 8081 is the worker port by default and binding it will block jobs. So
+      # we skip to 8082 instead of using the incremented port of 8081
+      SPARK_MASTER_WEBUI_PORT: 8082
```
jeancochrane: **[Question, non-blocking]** Do we actually need to change the port numbers that we bind within the container? My understanding is that the container's ports are fully isolated until we publish them to host ports using the `ports` service config, so there's no risk of port collision until we publish them to a host port. If that's so, can't we continue binding e.g. `SPARK_MASTER_WEBUI_PORT` to `8080` in the containers for both environments, and then just make sure to publish `8082:8080` in the `ports` config for the dev service?

jeancochrane (commented):
This is clever! I haven't tested it yet, but I had one high-level thought about the design that I want to make sure we discuss before I dig in further:

I trust that the profile approach works, and I look forward to testing it if we decide it's the correct path forward, but it strikes me that it's somewhat clunky insofar as we need to maintain duplicate definitions of services that share most of their logic, and users have to understand the profiles and how to use them correctly in the context of the compose CLI. I wonder if we could reduce duplication and cognitive overhead if we were to use [environment variable interpolation](https://docs.docker.com/compose/how-tos/environment-variables/variable-interpolation/) instead of profiles to adjust attributes that need to change between dev and prod environments?

By my count, it seems like we need to change the following Compose attributes between dev and prod:

- Root user ID (`services.{service}.user`)
- Image tag (`services.{service}.image`)
- Container name (`services.{service}.container_name`)
- Restart (`services.{service}.restart`)
- Service network name (`services.{service}.networks[0]`)
- Network name (`networks[0]`)
- Subnet address (`networks[0].ipam.config[0].subnet`)
- Host machine port numbers (`services.{service}.port`)
- Spark executor memory (`services.{service}.environment.SPARK_EXECUTOR_MEMORY`)
- Spark worker cores (`services.{service}.environment.SPARK_WORKER_CORES`)

Did I miss anything? If my understanding is correct, we should be able to modify each of these attributes by interpolating environment variables. We've already been using this approach to set the base URL for the image tag (`services.{service}.image`) to the `$CCAO_REGISTRY_URL` env var, but I expect we could use  it for the rest of the attributes too. Does that sound correct to you?

While I think environment variables would make for a cleaner configuration, I'm unsure how best to allow callers to manage their environments. As I see it, there are two ways we could keep dev and prod env vars separate:

* **Option 1**: Split the `.env` file into `.env.dev` and `.env.prod`, remove `.env` from version control, and update the README to instruct developers to copy `.env.dev` ➡️ `.env` when setting up their dev environment or copy `.env.prod` ➡️ `.env` when deploying changes to prod
  * **Pro**: Config is stored at the directory level, so developers don't need to think about it after they've completed setting up their environment
  * **Con**: Developers might miss the additional setup/deploy step
* **Option 2**: Split the `.env` file into `.env` (for development) and `.env.prod` (for prod), and update cron and prod docs to specify the use of the `--env-file` flag to point to `.env.prod`
  * **Pro**: No intervention necessary to set up a development environment
  * **Con**: Developers might forget to use the `--env-file` flag in prod

As far as I can tell, the main problem with both of these options is that we currently use the `.env` file to store secrets, so we would need a way of providing those secrets to callers without keeping them under version control in the dev or prod `.env` files. One way around this would be to switch to [the recommended secrets configuration](https://docs.docker.com/compose/how-tos/use-secrets/) to store our secrets instead of env vars. Seems to me that this would be a reasonable refactor to undertake regardless of the decision we make about how to separate dev and prod environments, although if we decide to move forward with profiles then I'm fine punting the secrets refactor to a later date.

I'm curious what you think about all this! I wouldn't be surprised if you've already thought about environment variables and decided against them for a good reason. Let me know if you want to chat about it on a call to speed up the feedback loop.

docker-compose.yaml
dfsnow: We don't! It was just left over from when I was merging multiple anchors (I'd split the env vars up into groups).

docker-compose.yaml
dfsnow: Fixed in the most recent revisions!

docker-compose.yaml
dfsnow: Good catch and TIL. Fixed in b8c81c5

docker-compose.yaml
dfsnow: You're right that we probably don't need to change them, but IMO it's better to play it safe here since Spark is pretty finnicky re: networking.

dfsnow:
@jeancochrane This is ready for another look. You were right, using env var interpolation is much simpler. I went with your option 2 and I think the whole thing is much simpler now.

Dockerfile
```diff
@@ -2,6 +2,7 @@ FROM bitnami/spark:3.5.1
 SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
 
 # Switch to root user so we can install jq and yq for parsing JSON/YAML
+# hadolint ignore=DL3002
```
dfsnow, line 5: GitHub superlinter gets mad about not having a non-root user specified in the Dockerfile.

jeancochrane: **[Question, non-blocking]** Remind me why we need to be root to install yq and jq?

README.md
```diff
@@ -192,62 +192,104 @@ populate the following:
...
 - `secrets/` - These are credential files needed to connect to other systems.
-- `.env` - This file sets a few non-critical but still private options.
+
+## Using the development environment
+
+The Docker Compose stack we use to run Spark (via `docker compose up -d`)
+has a separate development environment that can be used to test code changes
+without disrupting the production containers. The development environment is
+configured via a dedicated
+[env-file](https://docs.docker.com/compose/how-tos/environment-variables/variable-interpolation/#additional-information-1).
```
jeancochrane: **[Suggestion, non-blocking]** I wonder if a more direct link might be https://docs.docker.com/compose/how-tos/environment-variables/variable-interpolation/#env-file?

README.md
```diff
@@ -192,62 +192,104 @@ populate the following:
...
+docker exec -it spark-node-master-dev ./submit.sh \
+    --json-string "$(yq -o=json .test_jobs ./config/default_jobs.yaml)"
+```
+
+A typical development workflow might look something like:
+
+1. Clone the repository to your own machine or home directory. Do _not_ use
+  the production `shiny-server` copy of the repository for development.
+2. Copy the secrets and drivers from the production setup to the development
+  repository. See [Files not included](#files-not-included).
```
jeancochrane: **[Suggestion, non-blocking]** It might be nice to have an example of these commands -- all the other steps are pretty straightforward, but this one requires the reader to go looking for the files they want and then figure out how to copy them over correctly.

docker-compose.yaml
```diff
@@ -1,12 +1,12 @@
 x-default: &node
-  image: ${CCAO_REGISTRY_URL}/service-spark-iasworld:latest
   # Corresponds to the user set in the Dockerfile and the shiny-server
   # user on the server for proper write perms to mounted directories
-  user: 1003:0
+  user: "${UID:-1003}:0"
+  image: ghcr.io/ccao-data/service-spark-iasworld${ENV:+:}${ENV:?}
```
jeancochrane: **[Question, non-blocking]** Two questions about this:

1. Since question-mark expansion will raise an error if the parameter is not set (at least according to my reading of [the docs](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html)), do we need the conditional expansion of the colon via `${ENV:+:}`?
2. Any concerns about changing the prod image tag from `:latest` to `:prod`? Seems fine to me, but I just want to double-check that this is an intentional change.

docker-compose.yaml
```diff
@@ -19,70 +19,52 @@ x-default: &node
...
+    - IPTS_CONNECTION
+    - SPARK_ENV
   networks:
     - sparknet
 
+# These environmental variables get merged into the env vars of containers.
+# See: https://docs.docker.com/reference/compose-file/extension/#example-4
+x-spark-env: &spark-env
+  SPARK_MASTER_URL: spark://spark-node-master${ENV:+-}${ENV:?}:7077
+  SPARK_PUBLIC_DNS: ${SPARK_PUBLIC_IP}
```
jeancochrane: **[Suggestion, non-blocking]** I like that you updated the `$ENV` and `$SPARK_RESTART_POLICY` references above to use the parameter expansion syntax that raises an error if the env var is unset (`${var:?}`), I wonder if it's worth doing that for every reference to a required env var in this file?

src/submit_jobs.py
```diff
@@ -114,10 +114,9 @@ def submit_jobs(
     extract_target: str,
     json_file: str | None = None,
     json_string: str | None = None,
-    run_github_workflow: bool = True,
-    run_glue_crawler: bool = True,
-    upload_logs: bool = True,
```
jeancochrane, line 119: **[Question, non-blocking]** What's the reason for removing `upload_logs` as an option? If we're going to get rid of it, should we also remove the `--upload-logs` argument from the CLI?

README.md
```diff
@@ -102,15 +102,15 @@ For example, to submit the test jobs in `config/default_jobs.yaml` via
 `--json-string`, run the following command:
 
 ```bash
-docker exec spark-node-master ./submit.sh \
+docker exec spark-node-master-prod ./submit.sh \
```
jeancochrane: **[Question, non-blocking]** Have we thought about using [`docker compose exec`](https://docs.docker.com/reference/cli/docker/compose/exec/) instead of `docker exec` so that callers don't have to remember to reference the correct container name for their environment? Apologies if we've already discussed this -- I can't remember doing so!

jeancochrane (approved):
I love the refactor! A couple of small questions and nits, but nothing serious. Resolve how you see fit and feel free to re-request me if you want a final look at anything.

Dockerfile
dfsnow: We need to be root to install to the system library for yq/jq and for Python packages. Else we need to mess with faffy user-space packages and virtual envs. I don't think we technically need to set root here though, since it's the default.

README.md
dfsnow: Good idea! Switched in bea2ea7.

Doesn't work as well for the cron jobs though, since you'd have to `cd` to the dir and I'd rather have them explicitly target prod.

README.md
dfsnow: Fixed in db179ca!

src/submit_jobs.py
dfsnow: It's not used in the `submit_jobs()` call anymore. Instead it's used on L289 to determine whether or not to run a separate log upload process. We moved it there so logs still get uploaded if `submit_jobs()` fails.

docker-compose.yaml
dfsnow: 1. Nope, removed that and the expansion of `-` in 151d5b8
2. Nah, the image tag is totally arbitrary. It's fine to use `prod`.

docker-compose.yaml
dfsnow: Done in 151d5b8.

README.md
dfsnow: Added in c93facc.

README.md
jeancochrane: I think using `docker compose exec` for manual runs and `docker exec` for cron makes sense 👍🏻

README.md
jeancochrane: You might have caught this already, but I think we want `cp` instead of `mv` here: https://github.com/ccao-data/service-spark-iasworld/blob/c93faccc221d605081b9ac86efebd497dba88a42/README.md?plain=1#L224-L225

PR #23, opened 2024-11-18
Switch back to Compose profiles for development

dfsnow:
It turns out you can't use env var interpolation to spin up multiple copies (prod and dev) of the same service, since the copies will still share the same service name under the hood. As such, this PR reverts some of the changes from #22 by switching back to [Compose profiles](https://docs.docker.com/compose/how-tos/profiles/) for managing our `dev` and `prod` environments.


jeancochrane (approved):
Nice work! Bummer that env var interpolation doesn't work, but I'm happy with this approach too.

One note: Do you mind updating the [`export_qc_town_close_reports` script](https://github.com/ccao-data/data-architecture/blob/4314f7d014fcf8f4c5f149899612c91f016667b7/dbt/scripts/export_qc_town_close_reports.py#L184-L192) and [dbt README](https://github.com/ccao-data/data-architecture/blob/4314f7d014fcf8f4c5f149899612c91f016667b7/dbt/README.md?plain=1#L893-L898) in `data-architecture` to make sure that our tools and docs over there match these interface changes?

PR #24, opened 2024-11-19
Update Spark prod resources configuration

dfsnow:
I goofed and bumped the core count to 32 in #22. This results in OOM errors on our server. Lowering the core count to 28 prevents this.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -98,8 +98,8 @@ services:
       <<: *spark-env
       SPARK_MODE: worker
       SPARK_MASTER_URL: spark://spark-node-master-prod:7077
-      SPARK_WORKER_MEMORY: 96g
-      SPARK_WORKER_CORES: 32
+      SPARK_WORKER_MEMORY: 96G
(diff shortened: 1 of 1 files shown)


docker-compose.yaml
```diff
@@ -98,8 +98,8 @@ services:
       <<: *spark-env
       SPARK_MODE: worker
       SPARK_MASTER_URL: spark://spark-node-master-prod:7077
-      SPARK_WORKER_MEMORY: 96g
-      SPARK_WORKER_CORES: 32
+      SPARK_WORKER_MEMORY: 96G
```
jeancochrane, line 101: **[Question, non-blocking]** Did you determine whether the capitalization matters here or not?

jeancochrane (approved):
Thanks for troubleshooting this!

docker-compose.yaml
dfsnow: Based on super quick tests, I don't believe it does. I just reverted it to the way it was before to be safe.

PR #32, opened 2025-12-17
Split out config files for default jobs, and change `--json-file` arg to `--yaml-file`

jeancochrane:
This PR implements two changes in the script that runs the Spark pipeline:

- Splits out the batch job configs from the `config/default_jobs.yaml` file to dedicated files in a new subdirectory, `config/default_jobs/<batch_name>.yaml`
- Changes the `--json-file` argument to `--yaml-file`

The motivation for these changes is that `yq` suddenly stopped working in our cron jobs due to snap's [strict confinement](https://github.com/mikefarah/yq?tab=readme-ov-file#snap-notes), which requires a logged-in user session to exist in order for the snap package (in this case, `yq`) to run. See [this StackExchange question](https://askubuntu.com/questions/1515134/how-to-start-chromium-from-a-service-cron-for-nightly-website-verification) for an example of the same problem with a different snap package.

I'm not 100% what changed such that `yq` only just broke recently, since in theory it seems like it should always have been a problem. However, I confirmed the root cause by running the same cron job while logged in and logged out, and staying logged out consistently reproduced the error.

Since parsing the `default_jobs.yaml` config is the only task we use `yq` for in this service, we decided the simplest path forward was to split out the jobs into their own dedicated YAML files, and then update the `--json-file` script arg so that it could read YAML files and hence not require `yq` to convert the YAML to JSON.

This change is live on prod and has so far successfully run the `weekday_jobs.yaml` file and `test_jobs.yaml` file via cron while logged out. We will test `weekend_jobs.yaml` when it runs this weekend.


config/default_jobs/test_jobs.yaml
```diff
@@ -0,0 +1,26 @@
+# Jobs with a good mix of situations used for testing
```
jeancochrane, line 1: This diff only makes two changes to these config files in the process of splitting them out from the prior consolidated `default_jobs.yaml` file. I'll call out the changes in specific inline comments, but here's an overview of the changes:

1. Removes any `define` blocks that we were using to set YAML anchors (i.e. variables) for `min_year`, because otherwise Spark would interpret those blocks as table configs, and raise an error due to the table config being malformed.
    * This change only affects `test_jobs.yaml` and `weekday_jobs.yaml`, which were the only two job batches to use the `min_year` anchor in the old `default_jobs.yaml` file. In the case of `test_jobs`, I removed the `min_year` anchor entirely, since we were only using it for one table config. In the case of `weekday_jobs`, I moved the `min_year` anchor definition to the first occurence of the `min_year` in the file.
2. Removes the top-level key from the YAML definition (e.g. `test_jobs` in the case of this file) since it was only necessary in order to support selecting specific batch configs from the consolidate `default_jobs.yaml` file using `yq`.

config/default_jobs/test_jobs.yaml
```diff
@@ -0,0 +1,26 @@
...
+aasysjur:
+  table_name: iasworld.aasysjur
+  min_year: null
+  max_year: null
+  cur: null
+  predicates_path: null
+addn:
+  table_name: iasworld.addn
+  min_year: 2023
```
jeancochrane, line 10: This `min_year` attribute used to reference the `min_year` anchor set in a `define` block at the top of `default_jobs.yaml`, but I've removed that alias because this is the only table that references the anchor in this batch. Here's what the code looked like prior to this diff, for the sake of comparison:

https://github.com/ccao-data/service-spark-iasworld/blob/3fe41bc7df3360794cca6d4a932745c7f0870280/config/default_jobs.yaml#L1-L2

https://github.com/ccao-data/service-spark-iasworld/blob/3fe41bc7df3360794cca6d4a932745c7f0870280/config/default_jobs.yaml#L274-L276

config/default_jobs/weekday_jobs.yaml
```diff
@@ -0,0 +1,71 @@
+# Jobs that run daily at 1AM, typically getting the 2 most
+# recent years of data for critical tables
+addn:
+  table_name: iasworld.addn
+  # Update this year to update the rest of the minimum years in
+  # the file
+  min_year: &min_year 2023
```
jeancochrane, line 7: This diff moves the `min_year` anchor definition to the first occurrence in the file, rather than defining it in a dedicated `define`  block at the top of the file.

README.md
```diff
@@ -262,14 +268,11 @@ file below schedules daily jobs for frequently updated tables and weekly ones
 for rarely-updated tables.
 
 ```bash
-# Extract recent years from frequently used tables on weekdays at 1 AM CST
-0 6 * * 1,2,3,4,5 docker exec spark-node-master-prod ./submit.sh --json-string "$(yq -o=json .default_jobs /full/path/to/default_jobs.yaml)" --run-github-workflow --run-glue-crawler --upload-data --upload-logs
+# Extract recent years from frequently used tables on weekdays at 4 AM CST
```
jeancochrane, line 271: The timing of this job had drifted out of date; at some point in the past we changed it to 4am in the prod crontab, not 1am.

wrridgeway (approved):
I do understand what's going on here, but thankfully it seems like your tests are confirming everything is working as expected because I don't think I'm familiar enough with shell scripting to really audit this work.

Issue #31, opened 2025-11-17
Migrate away from deprecated Bitnami Spark Docker image


jeancochrane:
Sounds like [Bitnami is no longer publishing Spark images for free](https://hub.docker.com/r/bitnami/spark):

> This image is no longer available for free through Docker Hub. This image is available as a built OCI artifact in both Debian and Photon base OS formats through a commercial subscription of Bitnami Secure Images.

Boo. We need to switch to a different image, since now [our builds are failing due to the image not existing](https://github.com/ccao-data/service-spark-iasworld/actions/runs/19443076564/job/55630924596).

I don't have a lot of context into why we were using the Bitnami image in the first place, so I don't yet have a great idea of which direction we should go. Spark publishes [an official image](https://hub.docker.com/_/spark) but I feel like we probably would have chosen to use that one if it were easy, so I suspect there are some dragons lurking here that we don't know about.

Not urgent because the server still has access to the old Bitnami image, but this will block us from merging PRs until we can resolve it.

PR #33, opened 2025-12-17
Switch to legacy Bitnami Docker image, update Python dependencies, and remove `yq`/`jq`

jeancochrane:
This PR makes a few changes to fix our Spark Docker image, which [Bitnami has recently deprecated](https://github.com/bitnami/containers/issues/83267), and hence [causing our CI builds to fail](https://github.com/ccao-data/service-spark-iasworld/actions/runs/19443076564/job/55630924596):

- Changes the base image from `bitnami/spark` ➡️ `bitnamilegacy/spark`
    - [`bitnamilegacy`](https://hub.docker.com/u/bitnamilegacy) is the archive for the old `bitnami` images, which will not receive support going forward (Bitnami is making these images subscriber-only, see the GitHub discussion linked above for documentation of this change). In the medium-term, we need to switch to a better-supported image (or roll our own), so I'm leaving https://github.com/ccao-data/service-spark-iasworld/issues/31 open to track that effort. For now, however, the legacy images are the fastest way to get builds passing again.
- Updates Python dependencies, which were slightly out of date such that a few of them could not be installed anymore
- Removes `yq` and `jq` installation, which we don't use, and definitively will never need once #32 lands

Ways I've tested this change:

- Checked that builds are now passing
- Ran a test ingest on the server using the new image to confirm that it continues to work

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM bitnamilegacy/spark:3.5.6
+FROM bitnamilegacy/spark:3.5.1
 SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
 
 # hadolint ignore=DL3002
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,18 +1,7 @@
 FROM bitnamilegacy/spark:3.5.6
 SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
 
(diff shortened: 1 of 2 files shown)


wrridgeway (approved):
Thanks for handling this! It's annoying to have to use 4 yaml files rather than one, but maybe not needing additional dependencies is a nice consolation prize.

PR #34, opened 2026-03-19
Configure log upload function to upload logs regardless of error

### LINKED ISSUE #29: Logs are failing to upload to CloudWatch on encountering some errors
I was debugging some silent failures that we saw this weekend, and got this traceback:

```
14/10/25 15:18:19 ERROR __main__: An error occurred while calling o3491.parquet.
: org.apache.spark.SparkException: Job aborted due to stage failure: Master removed our application: KILLED
        at org.apache.spark.scheduler.DAGScheduler.failJobAndIndependentStages(DAGScheduler.scala:2856)
        at org.apache.spark.scheduler.DAGScheduler.$anonfun$abortStage$2(DAGScheduler.scala:2792)
        at org.apache.spark.scheduler.DAGScheduler.$anonfun$abortStage$2$adapted(DAGScheduler.scala:2791)
        at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:62)
        at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:55)
        at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:49)
        at org.apache.spark.scheduler.DAGScheduler.abortStage(DAGScheduler.scala:2791)
        at org.apache.spark.scheduler.DAGScheduler.$anonfun$handleTaskSetFailed$1(DAGScheduler.scala:1247)
        at org.apache.spark.scheduler.DAGScheduler.$anonfun$handleTaskSetFailed$1$adapted(DAGScheduler.scala:1247)
        at scala.Option.foreach(Option.scala:407)
        at org.apache.spark.scheduler.DAGScheduler.handleTaskSetFailed(DAGScheduler.scala:1247)
        at org.apache.spark.scheduler.DAGSchedulerEventProcessLoop.doOnReceive(DAGScheduler.scala:3060)
        at org.apache.spark.scheduler.DAGSchedulerEventProcessLoop.onReceive(DAGScheduler.scala:2994)
        at org.apache.spark.scheduler.DAGSchedulerEventProcessLoop.onReceive(DAGScheduler.scala:2983)
        at org.apache.spark.util.EventLoop$$anon$1.run(EventLoop.scala:49)
        at org.apache.spark.scheduler.DAGScheduler.runJob(DAGScheduler.scala:989)
        at org.apache.spark.SparkContext.runJob(SparkContext.scala:2398)
        at org.apache.spark.sql.execution.datasources.FileFormatWriter$.$anonfun$executeWrite$4(FileFormatWriter.scala:307)
        at org.apache.spark.sql.execution.datasources.FileFormatWriter$.writeAndCommit(FileFormatWriter.scala:271)
        at org.apache.spark.sql.execution.datasources.FileFormatWriter$.executeWrite(FileFormatWriter.scala:304)
        at org.apache.spark.sql.execution.datasources.FileFormatWriter$.write(FileFormatWriter.scala:190)
        at org.apache.spark.sql.execution.datasources.InsertIntoHadoopFsRelationCommand.run(InsertIntoHadoopFsRelationCommand.scala:190)
        at org.apache.spark.sql.execution.command.DataWritingCommandExec.sideEffectResult$lzycompute(commands.scala:113)
        at org.apache.spark.sql.execution.command.DataWritingCommandExec.sideEffectResult(commands.scala:111)
        at org.apache.spark.sql.execution.command.DataWritingCommandExec.executeCollect(commands.scala:125)
        at org.apache.spark.sql.execution.QueryExecution$$anonfun$eagerlyExecuteCommands$1.$anonfun$applyOrElse$1(QueryExecution.scala:107)
        at org.apache.spark.sql.execution.SQLExecution$.$anonfun$withNewExecutionId$6(SQLExecution.scala:125)
        at org.apache.spark.sql.execution.SQLExecution$.withSQLConfPropagated(SQLExecution.scala:201)
        at org.apache.spark.sql.execution.SQLExecution$.$anonfun$withNewExecutionId$1(SQLExecution.scala:108)
        at org.apache.spark.sql.SparkSession.withActive(SparkSession.scala:900)
        at org.apache.spark.sql.execution.SQLExecution$.withNewExecutionId(SQLExecution.scala:66)
        at org.apache.spark.sql.execution.QueryExecution$$anonfun$eagerlyExecuteCommands$1.applyOrElse(QueryExecution.scala:107)
        at org.apache.spark.sql.execution.QueryExecution$$anonfun$eagerlyExecuteCommands$1.applyOrElse(QueryExecution.scala:98)
        at org.apache.spark.sql.catalyst.trees.TreeNode.$anonfun$transformDownWithPruning$1(TreeNode.scala:461)
        at org.apache.spark.sql.catalyst.trees.CurrentOrigin$.withOrigin(origin.scala:76)
        at org.apache.spark.sql.catalyst.trees.TreeNode.transformDownWithPruning(TreeNode.scala:461)
        at org.apache.spark.sql.catalyst.plans.logical.LogicalPlan.org$apache$spark$sql$catalyst$plans$logical$AnalysisHelper$$super$transformDownWithPruning(LogicalPlan.scala:32)
        at org.apache.spark.sql.catalyst.plans.logical.AnalysisHelper.transformDownWithPruning(AnalysisHelper.scala:267)
        at org.apache.spark.sql.catalyst.plans.logical.AnalysisHelper.transformDownWithPruning$(AnalysisHelper.scala:263)
        at org.apache.spark.sql.catalyst.plans.logical.LogicalPlan.transformDownWithPruning(LogicalPlan.scala:32)
        at org.apache.spark.sql.catalyst.plans.logical.LogicalPlan.transformDownWithPruning(LogicalPlan.scala:32)
        at org.apache.spark.sql.catalyst.trees.TreeNode.transformDown(TreeNode.scala:437)
        at org.apache.spark.sql.execution.QueryExecution.eagerlyExecuteCommands(QueryExecution.scala:98)
        at org.apache.spark.sql.execution.QueryExecution.commandExecuted$lzycompute(QueryExecution.scala:85)
        at org.apache.spark.sql.execution.QueryExecution.commandExecuted(QueryExecution.scala:83)
        at org.apache.spark.sql.execution.QueryExecution.assertCommandExecuted(QueryExecution.scala:142)
        at org.apache.spark.sql.DataFrameWriter.runCommand(DataFrameWriter.scala:859)
        at org.apache.spark.sql.DataFrameWriter.saveToV1Source(DataFrameWriter.scala:388)
        at org.apache.spark.sql.DataFrameWriter.saveInternal(DataFrameWriter.scala:361)
        at org.apache.spark.sql.DataFrameWriter.save(DataFrameWriter.scala:240)
        at org.apache.spark.sql.DataFrameWriter.parquet(DataFrameWriter.scala:792)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:569)
        at py4j.reflection.MethodInvoker.invoke(MethodInvoker.java:244)
        at py4j.reflection.ReflectionEngine.invoke(ReflectionEngine.java:374)
        at py4j.Gateway.invoke(Gateway.java:282)
        at py4j.commands.AbstractCommand.invokeMethod(AbstractCommand.java:132)
        at py4j.commands.CallCommand.execute(CallCommand.java:79)
        at py4j.ClientServerConnection.waitForCommands(ClientServerConnection.java:182)
        at py4j.ClientServerConnection.run(ClientServerConnection.java:106)
        at java.base/java.lang.Thread.run(Thread.java:840)

Failed to upload log file to CloudWatch: time data 'org.apache.spark.SparkException:' does not match format '%Y-%m-%d_%H:%M:%S.%f'
```

This traceback seems to indicate that Spark cannot upload some error messages like the one above to CloudWatch due to an error in timestamp parsing. I'm guessing that the log line showing the traceback does not have a timestamp, which breaks one of our assumptions about how CloudWatch logs should look. I haven't investigated the problem yet, though.

I caused this error by killing the job in the Spark console, meaning that this problem should be easy to reproduce and verify a fix.

wrridgeway:
My python debugging is not good, so hopefully I haven't totally missed the mark here:

I spent a while looking at the logs trying to figure out why they weren't getting properly formatted if there was an error during the spark job: from what I could see the logs were fine, we were just triggering the exception that says we failed to upload logs without even trying.

Here's an example of the spark ingest script [uploading logs even after a failure](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/$252Fccao$252Fjobs$252Fspark/log-events/iasworld_prod_2026-04-06_19-21-48) with the new changes.

Edit: Also making the secrets folder easier to work with in this PR, by renaming our examples to [`.example` files](https://stackoverflow.com/a/52084653) and ignoring anything else in the secrets folder.


jeancochrane (commented):
Thanks for picking this up! This is not exactly the solution I was imagining -- apologies for not making that clearer up front.

The particular edge case we're trying to fix is the case where one or more log lines are missing a timestamp. In the motivating example, log lines are missing timestamps because they are part of a traceback, which Spark does not format with a leading timestamp the way it formats other lines. However, I don't think the current version of the code in this branch handles that case -- if it runs into a message with an invalid timestamp format (or missing timestamp) on line 96 in `line.split(" ", 1)`, it will skip the rest of the logs and jump straight to the `finally` block, only uploading the lines that it successfully parsed before encountering the line that was missing a timestamp. Hence, I would expect the process to continue to fail to upload the traceback (though it would upload the logs before that, which I do think is a slight improvement).

I'm pretty sure your test job didn't end up dumping a traceback, but I'm not totally sure why. Regardless, from a quick look at [the logs you shared](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/$252Fccao$252Fjobs$252Fspark/log-events/iasworld_prod_2026-03-19_15-42-20), it seems like every line already had a unique timestamp anyway, so I doubt we were actually testing the edge case at hand here.

In my opinion, the right approach here is the tweak the way we process the logs so that we catch every line that is missing a timestamp, bundling them with the timestamp of the preceding log line. This would require checking each log line for a timestamp with a regex like this:

```python
TIMESTAMP_RE = re.compile(
    r"^(\d{4}-\d{2}-\d{2}_\d{2}:\d{2}:\d{2}\.\d{6}) (.+)"
)
```

Then, we would refactor `upload_logs_to_cloudwatch()` so that as it is iterating each log line, it's keeping track of the previous timestamp, and bundling together any logs that are missing timestamps with the most recent preceding timestamp. Here's some vibe code demonstrating that idea -- I haven't tested it yet, but it looks directionally correct to me:

```python
        try:
            with open(log_file_path, "r") as log_file:
                log_events: list[dict[str, int | str]] = []
                current_timestamp: int | None = None
                current_lines: list[str] = []

                for line in log_file:
                    line = line.rstrip("\n")
                    match = TIMESTAMP_RE.match(line)
                    if match:
                        # Flush the previous event before starting a new one
                        if current_timestamp is not None:
                            log_events.append(
                                {
                                    "timestamp": current_timestamp,
                                    "message": "\n".join(current_lines),
                                }
                            )
                        current_timestamp = int(
                            datetime.strptime(
                                match.group(1), "%Y-%m-%d_%H:%M:%S.%f"
                            ).timestamp()
                            * 1000
                        )
                        current_lines = [match.group(2).strip()]
                    else:
                        # No timestamp: continuation line (e.g. a traceback)
                        if current_timestamp is not None:
                            # Append to the current event
                            current_lines.append(line.strip())
                        else:
                            # No preceding event yet -- use the current time
                            # as a best-effort timestamp
                            log_events.append(
                                {
                                    "timestamp": int(
                                        datetime.now().timestamp() * 1000
                                    ),
                                    "message": line.strip(),
                                }
                            )

                # Flush the final event
                if current_timestamp is not None:
                    log_events.append(
                        {
                            "timestamp": current_timestamp,
                            "message": "\n".join(current_lines),
                        }
                    )

            # Sort log events by timestamp
            log_events.sort(key=lambda event: event["timestamp"])

            # CloudWatch doesn't allow colon in stream names, so use a dash
            log_stream_name_fmt = log_stream_name.replace(":", "-")
            try:
                self.logs_client.create_log_stream(
                    logGroupName=log_group_name,
                    logStreamName=log_stream_name_fmt,
                )
            except self.logs_client.exceptions.ResourceAlreadyExistsException:
                pass

            self.logs_client.put_log_events(
                logGroupName=log_group_name,
                logStreamName=log_stream_name_fmt,
                logEvents=log_events,
            )
            print("Successfully uploaded log file to CloudWatch")

            # Remove the log file after successful upload
            os.remove(log_file_path)
            print(f"Successfully removed log file: {log_file_path}")

        except Exception as e:
            print(f"Failed to upload log file to Cloudwatch: {e}")
```

Does that make sense? I'm happy to talk this through out loud tomorrow, too.

wrridgeway:
Alright I'll try working through this. I was printing logs every time i killed the process (which was like 2 dozen times) and it never printed a line that didn't have a timestamp. I wonder if there's some way to better view that output that I'm not privy to.

src/utils/aws.py
```diff
@@ -89,26 +90,65 @@ def upload_logs_to_cloudwatch(
...
-                            "timestamp": timestamp,
-                            "message": message.strip(),
+                            "timestamp": current_timestamp,
+                            "message": "\n".join(current_lines),
                         }
                     )
 
+
+            # Remove lines with empty messages to avoid CloudWatch upload errors
+            log_events = [line for line in log_events if line.get("message")]
```
wrridgeway: I had to add this since it was trying to upload a final line with a timestamp and no message.

wrridgeway:
As far as I can tell your vibe-coded solution works just fine with the addition of a single line. New failure logs [here](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/$252Fccao$252Fjobs$252Fspark/log-events/iasworld_prod_2026-03-26_16-32-50).

src/utils/aws.py
```diff
@@ -89,26 +90,67 @@ def upload_logs_to_cloudwatch(
             log_stream_name: The name of the CloudWatch log stream to write to.
             log_file_path: The path to the log file to upload.
         """
+
+        # Declare the regex pattern for parsing timestamps at the start of log
+        # lines
+        TIMESTAMP_RE = re.compile(
+            r"^(\d{4}-\d{2}-\d{2}_\d{2}:\d{2}:\d{2}\.\d{6}) (.+)"
```
jeancochrane: **[Nitpick, required]** I think this regex doesn't actually match the timestamp format that Spark emits. You can see in the logs you shared that the timestamp that CloudWatch displays for each log line is different from the timestamp in the log message itself. In fact, CloudWatch is showing the same timestamp for every log line, which I assume is due to the fact that our script is not able to parse any timestamps from any of the log lines and so is defaulting to the time that `upload_logs_to_cloudwatch()` got called:

Image

A quick check in a [regex tester](https://regex101.com/) confirms that the issue for these particular log lines is that our regex pattern expects six decimals of precision, while Spark is only emitting three decimals. So I expect the following regex would work, assuming all Spark logs use the same timestamp format as the log lines that are visible in the logs you shared:


```suggestion
            r"^(\d{4}-\d{2}-\d{2}_\d{2}:\d{2}:\d{2}\.\d{3}) (.+)"
```

src/utils/aws.py
```diff
@@ -89,26 +90,67 @@ def upload_logs_to_cloudwatch(
...
+        )
+
         try:
             with open(log_file_path, "r") as log_file:
                 log_events: list[dict[str, int | str]] = []
+                current_timestamp: int | None = None
+                current_lines: list[str] = []
```
jeancochrane, line 109: **[Suggestion, optional]** The new design is somewhat complicated, so we might as well explain why we're doing it this way:

```suggestion
                # Initialize state variables for tracking log lines and
                # timestamps. We'll use these as we parse the logs in order to
                # assign timestamps for any log lines that are missing them --
                # in those cases, we use the most recent timestamp that
                # precedes the log line
                current_timestamp: int | None = None
                current_lines: list[str] = []
```

src/utils/aws.py
jeancochrane: **[Question, non-blocking]** Do you have a sense of where that empty message was coming from? Was there perhaps a log line at the end that was entirely a new line? I would love to figure out the cause of the empty final message before merging this code since it seems plausible that it might be pointing to a bug in our parsing logic. If you remove the `os.remove(log_file_path)` call on line 175 before testing this again, we should be able to get access to the raw log file and debug it.

src/utils/aws.py
```diff
@@ -89,26 +90,67 @@ def upload_logs_to_cloudwatch(
...
+                            "timestamp": current_timestamp,
+                            "message": "\n".join(current_lines),
                         }
                     )
 
+            # Remove any lines with empty messages to avoid CloudWatch upload
+            # errors
+            log_events = [line for line in log_events if line.get("message")]
             # Sort log events by timestamp
             log_events.sort(key=lambda event: event["timestamp"])
```
jeancochrane, line 161: **[Question, non-blocking]** Out of curiosity: Does this `sort` operation preserve the order of lines that have the same `timestamp` (namely, tracebacks)? I don't know enough Java to be able to read the Spark tracebacks so I can't tell if the example logs you shared are ordered properly.

jeancochrane (changes requested):
This is very close! Just one blocking bug fix and a couple of clarifying questions.

src/utils/aws.py
wrridgeway: So this shouldn't be an issue the way logging is currently working. It's _appending_ the traceback to one log entry, so it won't get sorted. See the log message that starts with `ERROR FileFormatWriter: Aborting job 2112021d-60e2-4162-9681-8cc7c0bb879f` or `ERROR __main__: An error occurred while calling o2942.pa` in [this](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/$252Fccao$252Fjobs$252Fspark/log-events/iasworld_prod_2026-04-06_19-21-48) log. I don't know if that's exactly what you're looking for, but it does seem to make sense to me as good way to log tracebacks.

jeancochrane: **[Praise, non-blocking]** I see what you mean now! Looks great, the UX is very clear in the CloudWatch console.

secrets/AWS_CREDENTIALS_FILE.example
jeancochrane, line 1: **[Thought, non-blocking]** A note for posterity: During the course of development Billy noticed that, while files in `secrets/` are ignored in `.gitignore`, these example files are force tracked, which means that maintainers will see diffs when copying prod secrets to their dev env per the README instructions. We decided to follow a common practice for `.env` files and append the example filenames with `.example` so that we can track them in version control without generating diffs when copying over prod files. See [this SO answer](https://stackoverflow.com/a/52084653) for more background on this pattern.

jeancochrane (approved):
Thanks for your persistence here, I'll sleep easier knowing this is fixed!

PR #36, opened 2026-04-09
Update lower bound year for daily job to 2025

jeancochrane:
We haven't updated the lower bound for our daily import job since 2024, and it's starting to show: The daily import job used to take 30 mins, and now it takes 1.5 hours. This has started to collide with the timing for downstream jobs, most notably the IC reference file export. This PR bumps our lower bound to 2025, since the daily job only ever needs to import the two most recent years of data.

### DIFF
--- a/config/default_jobs/weekday_jobs.yaml
+++ b/config/default_jobs/weekday_jobs.yaml
@@ -4,7 +4,7 @@ addn:
   table_name: iasworld.addn
   # Update this year to update the rest of the minimum years in
   # the file
-  min_year: &min_year 2023
+  min_year: &min_year 2025
 aprval:
   table_name: iasworld.aprval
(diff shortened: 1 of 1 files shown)


jeancochrane:
I only need @wrridgeway's approval here, but I'm tagging everyone else in data science for visibility into this change, since we've all been training up on this project recently.

wrridgeway (approved):
Thank you Jean! Do we need to add this bit of maintenance to any documentation anywhere?

jeancochrane:
> Thank you Jean! Do we need to add this bit of maintenance to any documentation anywhere?

Yup, I've got a ticket open for next sprint to document this stuff in a place where we will remember it: https://github.com/ccao-data/wiki/issues/68

PR #37, opened 2026-04-10
Replace super-linter with pre-commit hooks

jeancochrane:
Super-linter had a serious security vulnerability recently (see https://github.com/ccao-data/service-spark-iasworld/security/dependabot/2). I tried to upgrade it to a secure version in https://github.com/ccao-data/service-spark-iasworld/commit/101b7473ce163e438cde16dbc312722b6c3fc1fd*, but a bunch of new linters/formatters failed checks. That got me thinking: We don't really even need super-linter in this repo, since there's a relatively static set of languages that this project uses (most importantly Python, bash, and YAML).

Rather than upgrade super-linter and disable/fix the changes it has introduced between versions 6 and 8, this PR gets builds passing again by replacing super-linter with the recommended pre-commit hooks for bash and YAML. While we're here, we also upgrade our version of ruff pre-commit hooks, and fix a small line of Python that is failing under newer versions of ruff.

---

* I didn't intend to push directly to the main branch; somehow, our org permissions were set to allow org owners to bypass branch protection, so when I accidentally pushed to main, git didn't stop me like it usually does. I flipped that setting back so that I am no longer able to bypass the rules, but in the meantime, I figure it's easier to put up a new PR than try to edit the git history to remove this accidental commit.


src/utils/spark.py
```diff
@@ -411,7 +411,7 @@ def upload(self, aws: AWSClient) -> list[str]:
             self.logger.info(
                 (
                     f"Table {self.table_name} deleting files: "
-                    f"{', '.join(map(lambda p: p.as_posix(), s3_files_to_delete))}"
```
jeancochrane, line 414: This line caused a circular conflict between `ruff-check` (linter) and `ruff-fmt` (formatter): The formatter wanted to break the expression onto multiple lines, but the linter noticed that multiline f-strings are not supported in Python 3.9 without using triple-quotes. We could have fixed this by switching to triple-quotes, but instead I just simplified the map logic so that it fits within the line length limit. This seemed like the right solution because list comprehension (like `x for x in iterable`) is a more Pythonic approach to iteration than `map` with `lambda` anyway,.

.pre-commit-config.yaml
```diff
@@ -1,8 +1,16 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.7.2
+    rev: v0.15.10
     hooks:
-      - id: ruff
+      - id: ruff-check
```
jeancochrane, line 5: Somewhere in the past few versions, `ruff-pre-commit` switched from `ruff` to `ruff-check` for the linter hook ID. See [the `ruff-pre-commit` docs](https://github.com/astral-sh/ruff-pre-commit?tab=readme-ov-file#using-ruff-with-pre-commit) for confirmation.

.pre-commit-config.yaml
```diff
@@ -1,8 +1,16 @@
...
           - --fix
       - id: ruff-format
+  - repo: https://github.com/adrienverge/yamllint
+    rev: v1.38.0
+    hooks:
+      - id: yamllint
+  - repo: https://github.com/shellcheck-py/shellcheck-py
+    rev: v0.11.0.1
+    hooks:
+    -   id: shellcheck
```
jeancochrane, line 16: [`adrienverge/yamllint`](https://github.com/adrienverge/yamllint) and [`shellcheck-py/shellcheck-py`](https://github.com/shellcheck-py/shellcheck-py) are both [officially recommended by the pre-commit maintaners](https://pre-commit.com/hooks.html).

wrridgeway (approved):
Thank you Jean, this is great.

PR #39, opened 2026-06-09
Add permissions to allow `build-docker` workflow to push to GHCR

jeancochrane:
[Main branch builds are failing](https://github.com/ccao-data/service-spark-iasworld/actions/runs/26979610745/job/79615235499) due to the `build-docker` workflow not having permissions to push to the GitHub Container Registry:

> ERROR: failed to build: failed to solve: failed to push ghcr.io/ccao-data/service-spark-iasworld:latest: denied: installation not allowed to Write organization package

The workflow did not fail on the PR (#38) because it is configured to [skip pushing the image on PRs](https://github.com/ccao-data/service-spark-iasworld/blob/a4861b71a04d401c9d1340d47ba85b9d4e45d4de/.github/workflows/build-docker.yaml#L23).

This PR fixes the build error by granting the workflow the `packages: write` permission.

To be totally honest, I don't understand why this workflow was succeeding before the upgrades to our third-party Docker actions. However, the fact that it is now failing without the `packages: write` permission makes sense to me, and I'm confident that it's appropriate for us to set that permission, so I'm not too worried about the fact that I don't understand why builds weren't failing before.

I tested this change by temporarily tweaking the workflow definition to allow it to push to GHCR on PRs in [1e3d125](https://github.com/ccao-data/service-spark-iasworld/pull/39/commits/1e3d1250d4d74b6081b045c1c2b245b869ad2603), then confirmed that [the push succeeded](https://github.com/ccao-data/service-spark-iasworld/actions/runs/27219224466/job/80369255546?pr=39#step:6:282).

### DIFF
--- a/.github/workflows/build-docker.yaml
+++ b/.github/workflows/build-docker.yaml
@@ -11,6 +11,9 @@ name: docker-build
 jobs:
   docker-build:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write # Allow workflow to push images to GHCR
     steps:
       - name: Checkout
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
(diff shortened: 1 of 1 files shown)


.github/workflows/build-docker.yaml
```diff
@@ -11,6 +11,9 @@ name: docker-build
 jobs:
   docker-build:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write # Allow workflow to push images to GHCR
```
jeancochrane, line 16: See the docs on [Authenticating to package registries with repository-scoped permissions](https://docs.github.com/en/packages/managing-github-packages-using-github-actions-workflows/publishing-and-installing-a-package-with-github-actions#authenticating-to-package-registries-with-repository-scoped-permissions) for instructions indicating that this is the correct choice:

> You should set the permissions for this access token in the workflow file to grant read access for the `contents` scope and write access for the `packages` scope.

wrridgeway (approved):
Bizarre that it wasn't failing before. Thanks for sorting this out.

PR #40, opened 2026-06-09
Add log stream for dev environment

### LINKED ISSUE #35: Add dedicated log stream for dev environment
It's not great that the code is hardcoded to [push logs to the prod log stream](https://github.com/ccao-data/service-spark-iasworld/blob/5772d54e4cfbad4e6600456cd04ce1fba9bd2228/src/submit_jobs.py#L295) even if it's running in a dev environment. Since dev and prod logs are mixed up, our alarms are not necessarily accurate during periods when we are doing active dev work. Ideally, dev environments would push to a dedicated dev log stream, with its own alarms that are clearly labeled as development alarms.

I think this should be relatively straightforward to do with environment variables, which we could then set differently for [dev](https://github.com/ccao-data/service-spark-iasworld/blob/5772d54e4cfbad4e6600456cd04ce1fba9bd2228/docker-compose.yaml#L47-L52) and [prod](https://github.com/ccao-data/service-spark-iasworld/blob/5772d54e4cfbad4e6600456cd04ce1fba9bd2228/docker-compose.yaml#L81-L86). Setting the log group via env var should be easy; altering the log message via env var might be slightly harder, so I consider it a nice-to-have.

wagnerlmichael:
Test log write is present in the spark-dev cloudwatch group.

Closes #35

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,7 +263,7 @@ A typical development workflow might look something like:
 > CloudWatch logs: the development environment uploads logs to a dedicated log
 > group (`/ccao/jobs/spark-dev`), set via the `CLOUDWATCH_LOG_GROUP_NAME`
 > environment variable in `docker-compose.yaml`, so that development runs
-> don't write to the production logs.
+> and production runs have clearly separated log destinations.
 
 ## Scheduling
 
--- a/README.md
+++ b/README.md
@@ -263,7 +263,7 @@ A typical development workflow might look something like:
 > CloudWatch logs: the development environment uploads logs to a dedicated log
 > group (`/ccao/jobs/spark-dev`), set via the `CLOUDWATCH_LOG_GROUP_NAME`
 > environment variable in `docker-compose.yaml`, so that development runs
-> don't trip production alarms.
+> don't write to the production logs.
 
 ## Scheduling
 
--- a/README.md
+++ b/README.md
@@ -255,11 +255,15 @@ A typical development workflow might look something like:
   `$SERVER_IP:8080`.
 
 > [!WARNING]
-> The development environment shares the same targets as the production
-> environment. That means it will write to the same S3 bucket/CloudWatch log
-> group and trigger the same workflows/crawlers (though all these features are
-> disabled by default). As such, use this environment carefully. If you mess up
-> production data, you can run the production version of the code to re-fetch it.
+> The development environment shares most of the same targets as the production
+> environment. That means it will write to the same S3 bucket and trigger the
+> same workflows/crawlers (though all these features are disabled by default).
+> As such, use this environment carefully. If you mess up production data, you
+> can run the production version of the code to re-fetch it. The exception is
+> CloudWatch logs: the development environment uploads logs to a dedicated log
+> group (`/ccao/jobs/spark_dev`), set via the `CLOUDWATCH_LOG_GROUP_NAME`
+> environment variable in `docker-compose.yaml`, so that development runs
+> don't trip production alarms.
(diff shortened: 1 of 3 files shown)


docker-compose.yaml
```diff
@@ -84,6 +85,7 @@ services:
       SPARK_UI_PORT: 4040
       SPARK_EXECUTOR_MEMORY: 82g
       SPARK_MASTER_WEBUI_PORT: 8080
+      CLOUDWATCH_LOG_GROUP_NAME: /ccao/jobs/spark
```
wagnerlmichael, line 88: I believe the prod service is still writing to the main prod log group, should we just leave it or restart it once we merge this?

jeancochrane: **[Suggestion, non-blocking]** I say we restart it once we merge, so that the prod service is up to date with the main branch of the code. That way if anything goes wrong, we can continue to have confidence that the state of the code on the main branch is the same as prod, which is helpful for debugging.

README.md
```diff
@@ -255,11 +255,15 @@ A typical development workflow might look something like:
...
-> production data, you can run the production version of the code to re-fetch it.
+> The development environment shares most of the same targets as the production
+> environment. That means it will write to the same S3 bucket and trigger the
+> same workflows/crawlers (though all these features are disabled by default).
+> As such, use this environment carefully. If you mess up production data, you
+> can run the production version of the code to re-fetch it. The exception is
+> CloudWatch logs: the development environment uploads logs to a dedicated log
+> group (`/ccao/jobs/spark-dev`), set via the `CLOUDWATCH_LOG_GROUP_NAME`
+> environment variable in `docker-compose.yaml`, so that development runs
+> and production runs have clearly separated log destinations.
```
jeancochrane, line 266: **[Thought, non-blocking]** The new version of this warning is making me realize that the lack of a dev S3 bucket for this job poses a bit of a conceptual problem for the notion of "dev logging". If the user is running a job in the dev environment and passes the flags `--upload-data` and `--upload-logs` to the script, then the dev job will upload the data to the prod bucket, while uploading the logs to the dev log group. This doesn't seem exactly correct, because the prod data _has_ been updated, but downstream log consumers that are querying the prod logs (most importantly our alerts defined in [`service-alerts`](https://github.com/ccao-data/service-alerts/)) won't be able to see that.

I think the simplest resolution would be to put up a follow-up PR that configures the dev job to upload to a dev bucket and call an associated dev crawler. This work will be slightly more challenging than the dev logging work, but I don't actually think it will be that bad. If we push to a dev bucket with a dev crawler, then we can have full confidence that dev runs will be completely isolated from prod runs. If you think that work is also important, then I'll go ahead and open that issue and we can work on it next sprint.

jeancochrane (approved):
Dev logs look great in the CloudWatch console! Nice work getting your head around this project.

This project doesn't have continuous deployment configured, because it's deployed on our server, but I figure we can pair on the manual deployment so you can learn how it works. Let me know when you're ready to merge and we can do that together.

README.md
wagnerlmichael: Good point! I'm assuming you've used the dev version of here more than anyone on the team. Do you think the workflows you've used would benefit from a separate dev bucket/crawler?

If so, I'm happy to continue to work on it. More work here seems like an excellent cross-training opportunity

jeancochrane: Great! I opened up https://github.com/ccao-data/service-spark-iasworld/issues/41 and dropped it into next sprint.

